PR #22, opened 2022-01-20
rbd hosts last

insatomcat:
To avoid the system thinking "rbd" is the official hostname of the host.


dupremathieu (changes requested):
Please add more details in your commit message.
Test: Ok

insatomcat:
it's done as you requested @dupremathieu

dupremathieu (approved):

PR #51, opened 2022-09-21
[debian] sysctl: change printk kernel parameters

insatomcat:
Signed-off-by: Florent CARLI <florent.carli@rte-france.com>

### DIFF
--- a/src/debian/00-schedrt.conf
+++ b/src/debian/00-schedrt.conf
@@ -1,1 +1,3 @@
 kernel.sched_rt_runtime_us=-1
+kernel.printk=3 3 3 3
+kernel.printk_devkmsg=off


src/debian/00-schedrt.conf
```diff
@@ -1 +1,3 @@
 kernel.sched_rt_runtime_us=-1
+kernel.printk=3 3 3 3
```
dupremathieu, line 2: You can only change the current printk level. So, it is "kernel.printk=3" not "3 3 3 3".

watare (approved):

src/debian/00-schedrt.conf
insatomcat: It seems to be working here:
```
root@virtu-elabo3:~# sysctl kernel.printk
kernel.printk = 7	4	1	7
root@virtu-elabo3:~# sysctl kernel.printk='3 3 3 3'
kernel.printk = 3 3 3 3
root@virtu-elabo3:~# sysctl kernel.printk
kernel.printk = 3	3	3	3
```

PR #69, opened 2022-10-31
[debian] VM template: pc-q35-3.1 + vsock

insatomcat:
Signed-off-by: Florent CARLI <florent.carli@rte-france.com>

### DIFF
--- a/templates/vm/votp_vm_rt_isolated.xml.j2
+++ b/templates/vm/votp_vm_rt_isolated.xml.j2
@@ -11,7 +11,7 @@
     <currentMemory unit="MiB">256</currentMemory>
     <vcpu placement="static">1</vcpu>
     <os firmware="efi">
-        <type arch="x86_64" machine="pc-i440fx-4.1">hvm</type>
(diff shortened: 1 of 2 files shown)


templates/vm/votp_vm_rt_isolated.xml.j2
```diff
@@ -61,7 +61,11 @@
             <target type="serial" port="0" />
         </console>
         <memballoon model="virtio">
+	    <stats period="5" />
         </memballoon>
+	<vsock model='virtio'>
+            <cid auto='yes' />
```
ycongal-smile, line 67: Maybe we will need to force this ID to be able to specify which VSOCK does what from the guest point-of-view.

insatomcat: the guest can only have one vsock anyway... as long as this "auto" approach works and does not create problem, I'm in favor of keeping it.
Just so you know I tested with several guests on the same host and it also works.

watare (approved):

PR #110, opened 2023-01-13
Add github action that call the ci on debian-main

eroussy:
Add the github action that calls the ci and configure cukinia and playbooks to run with it


.github/workflows/ci-debian.yml
```diff
@@ -0,0 +1,10 @@
+name: CI debian
```
dupremathieu, line 1: Add the license header and the copyright

dupremathieu (changes requested):

dupremathieu (approved):

PR #139, opened 2023-02-14
Anssi bp 028 R58

deribaucourt:


dupremathieu (approved):

roles/debian-hardening/tasks/main.yml
```diff
@@ -356,6 +356,17 @@
...
+- name: Uninstall apt packages not required after installation
+  apt:
+    name:
+      - linux-image-amd64
+      - linux-image-[!rt]-amd64
+      - fdisk
+      - ifupdown
+    state: absent
+    purge: yes
+    autoremove: yes
```
dupremathieu, line 368: Add when: not revert

dupremathieu (changes requested):

dupremathieu (approved):

src/debian/ssh-audit_hardening.conf.j2
```diff
@@ -32,4 +32,5 @@ Compression no
 ClientAliveInterval 300
 ClientAliveCountMax 0
 PermitTunnel no
-PermitOpen none
\ No newline at end of file
+PermitOpen none
+ListenAddress {{ ansible_env.SSH_CONNECTION.split(' ')[2] }}
```
dupremathieu, line 36: You cannot use this address. We do not know which address the SEAPATH user wants to bind sshd.
For me, this information should be described in the Ansible inventory.

roles/debian-hardening/tasks/main.yml
```diff
@@ -178,6 +181,14 @@
     state: absent
   when: revert
 
+- name: Wait for DHCP for SSH
```
dupremathieu, line 184: Wait for IP set
Because you can have a static IP (this is the case in the RTE CI)

roles/debian-hardening/tasks/main.yml
```diff
@@ -356,6 +367,28 @@
   register: services_to_restart
   with_items: "{{ hardened_services }}"
 
+- name: Uninstall apt packages not required after installation
+  apt:
+    name:
+      - fdisk
```
dupremathieu, line 373: Is it relative to BP-28 R58?

dupremathieu (changes requested):

roles/debian-hardening/tasks/main.yml
deribaucourt: Commit moved out of this PR.

src/debian/ssh-audit_hardening.conf.j2
deribaucourt: Commit moved out of this PR.

dupremathieu (approved):

PR #141, opened 2023-02-15
Anssi bp 028 r9

lucbeaufils:


src/debian/90-sysctl-hardening.conf
```diff
@@ -21,7 +21,7 @@ kernel.dmesg_restrict = 1
 kernel.kptr_restrict = 2
 
 # Only allow CAP_SYS_ADMIN processes to access "perf" subsystem
```
deribaucourt, line 23: Comment may not be accurate anymore.
Check if 3 wasn't a more secure option.

deribaucourt (changes requested):
Commit message is too generic "seapath: add kernel config".
I sugest "debian: hardening: sysctl: improve security"

src/debian/90-sysctl-hardening.conf
lucbeaufils: following this document : https://www.kernel.org/doc/Documentation/sysctl/kernel.txt
the highest secure option is ">=2: Disallow kernel profiling by users without CAP_SYS_ADMIN"
2 is specified by bp-028

deribaucourt (approved):

PR #150, opened 2023-02-21
Latency tests deployment

Paullgk:
This PR adds modification needed to run latency tests on cluster.
Following files and directories are added:

- `playbooks/test_run_latency_tests.yaml`: a playbook to run latency test in the cluster, using a dedicated publisher machine, and successively hypervisor and VM as subscriber machine. 
**Caution:** To deploy publisher tests, you will need a dedicated machine hosted as `standalone` in your inventory.
- `src/trafgen/` and `src/svtools`: source directories for publisher and subscriber source code.
- `tests_results/`: data results of publisher and subscriber tests directory.

### DIFF
--- a/playbooks/test_run_latency_tests.yaml
+++ b/playbooks/test_run_latency_tests.yaml
@@ -1,0 +1,150 @@
+# Copyright (C) 2023, RTE (http://www.rte-france.com)
+# SPDX-License-Identifier: Apache-2.0
+
+# This playbook run latency tests on a standalone publisher machine, with
+# successively hypervisor and VM as subscriber machine.
+
+---
+
+# First, starting hypervisor subscriber
+- hosts: hypervisors
+  name: Hypervisor subscriber tests
+  vars:
+    - svtools_archive_commit_id: 0c7b539
+    - poll_timeout: 5
+    - docker_test_image_name: hyp_sub_tests
+  tasks:
+  - name: Copy svtools archive on target
+    copy:
+      src: ../src/svtools/svtools_{{ svtools_archive_commit_id }}.tar
+      dest: /tmp/
+  - name: Load svtools as Docker image
+    command:
+      docker load --input /tmp/svtools_{{ svtools_archive_commit_id }}.tar
+  - name: Run subscriber tests
+    command:
+      docker run --name {{ docker_test_image_name }} -d --network host --rm
+      -it --cgroup-parent machine-rt.slice --cap-add=sys_nice
+      --cpuset-cpus="0-11" -v /data/:/root/ svtools svtools_subscriber_light
+      -i eth0 -o /root/svtools_out
+
+# Then starting standalone publisher
+- hosts: standalone
+  name: Publisher tests
+  vars:
+    - trafgen_archive_commit_id: a8936ce
+    - poll_timeout: 10
+    - docker_test_image_name: pub_tests
+    - network_interface: dk_8303
+    - network_device: eth0
+    - cpu_set: "2-6"
+  tasks:
+  - name: Copy trafgen archive on target
+    copy:
+      src: ../src/trafgen/trafgen_{{ trafgen_archive_commit_id }}.tar
+      dest: /tmp/
+  - name: Load trafgen as Docker image
+    command:
+      docker load --input /tmp/trafgen_{{ trafgen_archive_commit_id }}.tar
+  - name: Run publisher tests
+    command:
+      docker run --cap-add=sys_nice
+      --cpuset-cpus={{ cpu_set }} --privileged --network {{ network_interface }} --rm -v
+      /tmp/trafgen:/tmp/ trafgen --dev {{ network_device }} --conf /sv_template.cfg -n
+      100000 -t 10us -H
+
+  - name: Get publisher hypervisor tests results
+    fetch:
+      flat: yes
+      src: /tmp/trafgen/out
+      dest: ../tests_results/pub_results_hypervisor_{{ inventory_hostname }}
+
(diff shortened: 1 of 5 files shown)


playbooks/test_run_latency_tests.yaml
```diff
@@ -0,0 +1,150 @@
...
+    copy:
+      src: ../src/svtools/svtools_{{ svtools_archive_commit_id }}.tar
+      dest: /tmp/
+  - name: Load svtools as Docker image
+    command:
+      docker load --input /tmp/svtools_{{ svtools_archive_commit_id }}.tar
+  - name: Run subscriber tests
+    command:
+      docker run --name {{ docker_test_image_name }} -d --network host --rm
+      -it --cgroup-parent machine-rt.slice --cap-add=sys_nice
```
deribaucourt, line 27: do you need -it? It's not supposed to be interactive I think.

playbooks/test_run_latency_tests.yaml
```diff
@@ -0,0 +1,150 @@
...
+    copy:
+      src: ../src/trafgen/trafgen_{{ trafgen_archive_commit_id }}.tar
+      dest: /tmp/
+  - name: Load trafgen as Docker image
+    command:
+      docker load --input /tmp/trafgen_{{ trafgen_archive_commit_id }}.tar
+  - name: Run publisher tests
+    command:
+      docker run --cap-add=sys_nice
+      --cpuset-cpus={{ cpu_set }} --privileged --network {{ network_interface }} --rm -v
```
deribaucourt, line 52: Is --privileged required?

playbooks/test_run_latency_tests.yaml
```diff
@@ -0,0 +1,150 @@
...
+      regexp: '\n.*$'
+      replace: ''
+
+  - name: Get hypervisor subscriber tests results
+    fetch:
+      flat: yes
+      src: /data/svtools_out
+      dest: ../tests_results/sub_results_hypervisor_{{ inventory_hostname }}
+
+# Next, we do the same but with as VM as subscriber
```
deribaucourt, line 82: with a VM

playbooks/test_run_latency_tests.yaml
```diff
@@ -0,0 +1,150 @@
...
+    copy:
+      src: ../src/svtools/svtools_{{ svtools_archive_commit_id }}.tar
+      dest: /tmp/
+  - name: Load svtools as Docker image
+    command:
+      docker load --input /tmp/svtools_{{ svtools_archive_commit_id }}.tar
+  - name: Run subscriber tests
+    command:
+      docker run --name {{ docker_test_image_name }} -d --network host --rm
+      -it --cgroup-parent machine-rt.slice --cap-add=sys_nice
```
deribaucourt, line 100: -it required?

playbooks/test_run_latency_tests.yaml
```diff
@@ -0,0 +1,150 @@
...
+      --cpuset-cpus="0-11" -v /data/:/root/ svtools svtools_subscriber_light
+      -i eth0 -o /root/svtools_out
+
+# Then starting standalone publisher
+- hosts: standalone
```
deribaucourt, line 105: I guess it's not easy, isn't there a way to reuse the previous code with the hypervisor as publisher instead of duplicating it?

playbooks/test_run_latency_tests.yaml
```diff
@@ -0,0 +1,150 @@
...
+    copy:
+      src: ../src/trafgen/trafgen_{{ trafgen_archive_commit_id }}.tar
+      dest: /tmp/
+  - name: Load trafgen as Docker image
+    command:
+      docker load --input /tmp/trafgen_{{ trafgen_archive_commit_id }}.tar
+  - name: Run publisher VM tests
+    command:
+      docker run --cap-add=sys_nice
+      --cpuset-cpus={{ cpu_set }} --privileged --network {{ network_interface }} --rm -v
```
deribaucourt, line 125: --privileged?

src/trafgen/README
```diff
@@ -0,0 +1,8 @@
+trafgen docker archive belongs here.
```
deribaucourt, line 1: the git commit message only mentions gitignore, but I think the README part is more important. Proposition: "trafgen: document how to generate docker archive"
DO the same with the svtools commit.

deribaucourt (changes requested):

playbooks/test_run_latency_tests.yaml
Paullgk: I took docker command based on Thibaut recommendation, so I would prefer to not change that currently.

playbooks/test_run_latency_tests.yaml
Paullgk: I took docker command based on Thibaut recommendation, so I would prefer to not change that currently.

playbooks/test_run_latency_tests.yaml
Paullgk: I took docker command based on Thibaut recommendation, so I would prefer to not change that currently.

playbooks/test_run_latency_tests.yaml
Paullgk: I took docker command based on Thibaut recommendation, so I would prefer to not change that currently.

playbooks/test_run_latency_tests.yaml
Paullgk: It would be a nice improve, but currently we do prefer to keep a functional version of the playbook, and then make the optimization.

eroussy (approved):

PR #157, opened 2023-02-23
debian: hardening: configure grub password (BP-028-R5)

lucbeaufils:


roles/debian-hardening/tasks/main.yml
```diff
@@ -41,6 +41,18 @@
   with_items: "{{ kernel_params }}"
   when: revert
 
+- name: "Set grub password"
```
deribaucourt, line 44: put the ansible task and script in the same commit as they are strongly linked.

roles/debian-hardening/tasks/main.yml
```diff
@@ -41,6 +41,18 @@
...
+  copy:
+    src: "../src/debian/01_password"
+    dest: "/etc/grub.d/01_password"
+    mode: 0755
```
deribaucourt, line 48: set owner and group (root I guess)

src/debian/01_password
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/sh
+
+# Set grub password : "grub-mkpasswd-pbkdf2 -c 65536 -s 256 -l 64" and replace $GRUB_PASS
+# Default password : toor
```
deribaucourt, line 4: Not secure at all since this source code is public :stuck_out_tongue: 
OK for this commit, but add another one after that that takes the grub password from the inventory definition. Also document it.

src/debian/01_password
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/sh
+
+# Set grub password : "grub-mkpasswd-pbkdf2 -c 65536 -s 256 -l 64" and replace $GRUB_PASS
```
deribaucourt, line 3: Very nice to add this comment ! However, GRUB_PASS is missing

src/debian/01_password
```diff
@@ -0,0 +1,10 @@
...
+# Default password : toor
+
+GRUB_PASS=grub.pbkdf2.sha512.65536.5CFD987412BE492E5B25B524DA4FF8687841D62DD31F4CCE0608796204FAF7CF741D90B6DAB3F547C9625D995786F054836A060D45F96E64642CF029E2AF0503E030A8712DC71315F4C64E0D42CDC292CE6D8FBE63575EA4B1111AD42D61A89A406BAC02B47BA9C6BE9C7781E3ECBA3C506CC9B6020C3B808C85689687DB70C6E245E380A33F9ACF197FE9BD3C8FDA30588C429DBE4CD0FA022229C42E605C5DAF6E654AE0E160864DA2B6FDAA35247D88C886A876F402C2DDB56612F469A28F72551FD97C061E9630DCB8655A072DE0BA761D576F110C04E65DC3AE6EC6A62A89AA683934FEAE50EA0DF9BF6595CA4C9A3C1A1C11A657CCC40359CF7C6EB3E7.82DB321EBD4A971F73F49B4BEA96A42C47809C8E0E08D605A59C3460ECAF07ED199401A368EAA3787702F33A09795F5C6AC9CA12219229F890C16535F9A0544F
+
+echo superusers="root"
+echo password_pbkdf2 root $GRUB_PASS
+awk "/^menuentry 'Debian/,/}/" /boot/grub/grub.cfg | sed "s/menuentry.*{/menuentry 'SEAPATH' --unrestricted {/g"
```
deribaucourt, line 10: I think this could belong in build_debian_iso. This awk/sed patch is tricky and could fail if grub.cfg is not exactly as you want. Check if it wouldn't be more convenient and robust in build_debian_iso.

deribaucourt (changes requested):

src/debian/01_password
lucbeaufils: Use only sed, okey in this repo, Rename seapath

roles/debian-hardening/vars/main.yml
```diff
@@ -20,3 +20,6 @@ hardened_services:
   - pacemaker
   - syslog-ng
   - ceph-crash
+
+grub_password: grub.pbkdf2.sha512.65536.401EFA213B7334C4E6975DE70AEF0A61B35F64AD5516C0B111083DC7D067C73634C017E44B4B9BE65AAE74F61ACD31A3D52BCD3FE5FBF7BDFE6431C60D799BF762EBB5AC98C47839386E0503F626BD7BDA185EE3C64B1A77B630577C6CB333C5D222E943F8EFB6E4846CDADC1AEDF6B9DEE1EAEEAA606333B87AA14D6CF60D4A6D9CC3CF69C1F88EC86596A2B50174EB1A844A296B6950D6409A758FBF097B72E41736ADF1D8F3C3CAC42A3CB479A5DDABE2B08B7C4634691FCBCAB569596FE1B97C6F6085EC7D698C31C2C4350779B78DE445D12B02279FE6DD92AF1A66C48A7E88030865DEE9AC7F2D339774A1494ED3E025B9C8ECAFE9D9FECA54CA46F3F3.CCB479FE2FF11614F2CBDD687B467BB776B0B317543797C59AC06897975441433F83E332A8F72B9DE96FEDB27E5E01C98663576637E6E045759DB438DD47B2FC
```
deribaucourt, line 24: As discussed, this shouldn't be hardcoded. Add and document an ansible variable to the inventories.

src/debian/01_password
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/sh
+
+GRUB_PASS=
```
deribaucourt, line 3: Empty?

src/debian/01_password
```diff
@@ -0,0 +1,7 @@
...
+
+echo superusers="root"
+echo password_pbkdf2 root $GRUB_PASS
+awk "/^menuentry 'Debian/,/}/" /boot/grub/grub.cfg | awk '{gsub(/menuentry.*{/,"menuentry 'Debian' --unrestricted {"); print}'
```
deribaucourt, line 7: Could probably be inside a single command:
```
awk '/^menuentry 'Debian/{gsub(/menuentry.*{/,"menuentry 'Debian' --unrestricted {"); print}'
```

deribaucourt: Add a comment showing before/after command results since it's hard to decypher.
bad menuentry { --> new menuentry {

deribaucourt (changes requested):

src/debian/01_password
lucbeaufils: the GRUB_PASS is modified with ansible to permit usage of a ansible var

deribaucourt: Use jinja to explicit that this is a template file. This is more ansible-y than runing sed/awk commands.

roles/debian-hardening/templates/01_password.j2
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/sh
+
+{% if grub_password is defined %}
```
deribaucourt, line 3: grub_password shouldn't be optional.

roles/debian-hardening/tasks/main.yml
```diff
@@ -382,5 +381,22 @@
...
+    owner: "root"
+    mode: 0755
+  when: not revert
+- name: Unset grub password
+  file:
+    state: absent
+    path: /etc/grub.d/01_password
+  when: revert
+
+- name: Update grub
```
deribaucourt, line 398: rather than re running update-grub, move Set grub password before the last update-grub

deribaucourt: I remember that we've seen the CI doesn't rollback the boot partition.
It would be a good idea now to run update-grub in the rollback playbook after reboot.
Please add a commit for that.

deribaucourt (changes requested):

roles/debian-hardening/tasks/main.yml
lucbeaufils: I change the 01_password.j2 to override kernel boot parameter

playbooks/ci_restore_snapshot.yaml
```diff
@@ -20,3 +20,5 @@
         lv: root
         snapshot: root-snap
         size: 7G
+    - name: Update grub
```
deribaucourt, line 23: Would make more sense right after the reboot

deribaucourt (changes requested):

deribaucourt (approved):

PR #159, opened 2023-02-23
Fix cluster deployment

deribaucourt:


playbooks/cluster_setup_network.yaml
```diff
@@ -190,4 +190,3 @@
           wait_for_connection:
       when:
         - need_reboot is defined and need_reboot
-        - skip_reboot_setup_network is not defined or not skip_reboot_setup_network
```
dupremathieu, line 193: I suggest to keep this check. skip_reboot_setup_network can be set if needed in the inventory.

dupremathieu (changes requested):

dupremathieu (approved):

PR #161, opened 2023-02-24
Anssi bp 028 r73

Paullgk:
This PR adds a playbook  `playbooks/cluster_setup_auditd.yaml` to properly configure auditd on target:

- Redirect all auditd output to syslog
- Configure auditd rules
- Python scripts to generate syscall needed by Cukinia to test auditd rules

### DIFF
--- a/playbooks/test_deploy_cukinia_tests.yaml
+++ b/playbooks/test_deploy_cukinia_tests.yaml
@@ -18,8 +18,8 @@
         mode: 0755
     - name: Copy Cukinia's includes
       copy:
-        src: ../src/cukinia-tests/includes/kernel_config_functions
-        dest: /usr/share/cukinia/includes/kernel_config_functions
+        src: ../src/cukinia-tests/includes/
+        dest: /usr/share/cukinia/includes/
     - name: Create /usr/share/testdata
       file:
         path: /usr/share/testdata
--- a/roles/debian-hardening/tasks/main.yml
+++ b/roles/debian-hardening/tasks/main.yml
@@ -403,6 +403,16 @@
     path: /etc/grub.d/01_password
   when: revert
 
+- name: copy syslog.conf
+  copy:
+    src: ../src/debian/syslog.conf
+    dest: /etc/audit/plugins.d/syslog.conf
+
+- name: copy audit.rules
+  copy:
+    src: ../src/debian/audit.rules
+    dest: /etc/audit/rules.d/audit.rules
+
 - name: Update grub
   command: update-grub
 
--- a/roles/debian-hardening/tasks/main.yml
+++ b/roles/debian-hardening/tasks/main.yml
@@ -405,12 +405,12 @@
 
 - name: copy syslog.conf
   copy:
-    src: ../src/debian/syslog.conf
+    src: ../src/debian/auditd/syslog.conf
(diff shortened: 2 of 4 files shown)


src/debian/auditd/ioperm.py
```diff
@@ -0,0 +1,25 @@
+# Copyright (C) 2023, RTE (http://www.rte-france.com)
```
deribaucourt, line 1: I think these files are only used for testing and should rather be moved to the cukinia-tests repository.
We don't want testing executables to be deployed on a production system for cybersecurity.

playbooks/cluster_setup_auditd.yaml
```diff
@@ -8,10 +8,25 @@
...
+      src: ../src/debian/auditd/syslog.conf
       dest: /etc/audit/plugins.d/syslog.conf
 
   - name: copy audit.rules
     copy:
-      src: ../src/debian/audit.rules
+      src: ../src/debian/auditd/audit.rules
       dest: /etc/audit/rules.d/audit.rules
+
+  - name: copy python ioperm
```
deribaucourt, line 19: As stated in a previous comment, I think this should be moved to cukinia-tests

playbooks/cluster_setup_debian.yaml
```diff
@@ -10,6 +10,7 @@
 - import_playbook: cluster_setup_libvirt.yaml
 - import_playbook: cluster_setup_ha.yaml
 - import_playbook: cluster_setup_keys.yaml
+- import_playbook: cluster_setup_auditd.yaml
```
deribaucourt, line 13: should be called in cluster_setup_hardened_debian because this is part of hardening.

playbooks/cluster_setup_auditd.yaml
```diff
@@ -0,0 +1,17 @@
+# Copyright (C) 2023, RTE (http://www.rte-france.com)
```
deribaucourt, line 1: I think this would make more sense to be integrated in the debian hardened role.

deribaucourt (changes requested):

src/debian/auditd/ioperm.py
Paullgk: I moved them in Cukinia repository, see the commit: https://github.com/seapath/cukinia-tests/pull/30/commits/ab895ce098ebfecdd8787858b7bdd189303797ea

playbooks/cluster_setup_auditd.yaml
Paullgk: Done: https://github.com/seapath/cukinia-tests/pull/30/commits/ab895ce098ebfecdd8787858b7bdd189303797ea

playbooks/cluster_setup_debian.yaml
Paullgk: All tasks have been moved in debian hardening role.

playbooks/cluster_setup_auditd.yaml
Paullgk: All tasks have been moved in debian hardening role.

deribaucourt (changes requested):
Bump cukinia-tests

deribaucourt (approved):

PR #166, opened 2023-03-02
debian: hardening: fix memory and cpus vulnerabilities

lucbeaufils:


deribaucourt (approved):

roles/debian-hardening/vars/main.yml
```diff
@@ -7,11 +7,15 @@ kernel_params:
  - init_on_free=1
  - slab_nomerge
  - pti=on
- - slub_debug=ZF
+ - slub_debug=FZP
```
dupremathieu, line 10: Are you sure we need to enable slub poisoning?
According https://kernsec.org/wiki/index.php/Kernel_Self_Protection_Project/Recommended_Settings slug poisoning is not required if init_on_alloc=1
init_on_free=1 is enabled.

roles/debian-hardening/vars/main.yml
```diff
@@ -7,11 +7,15 @@ kernel_params:
...
  - randomize_kstack_offset=on
  - slab_common.usercopy_fallback=N
  - iommu=pt
  - security=yama
+ - page_poison=on
```
dupremathieu, line 15: Same thing as above page_poison=on is not mandatory with init_on_free=1.

roles/debian-hardening/vars/main.yml
```diff
@@ -7,11 +7,15 @@ kernel_params:
...
+ - mce=0
```
dupremathieu, line 16: Why disabling machine checking?
According to the documentation, the correct value to disable it is "off" not "0".

roles/debian-hardening/vars/main.yml
```diff
@@ -7,11 +7,15 @@ kernel_params:
...
+ - page_alloc.shuffle=1
```
dupremathieu, line 17: Already enabled in the kernel configuration
grep CONFIG_SHUFFLE_PAGE_ALLOCATOR config-6.0.0-0.deb11.6-rt-amd64 
CONFIG_SHUFFLE_PAGE_ALLOCATOR=y

dupremathieu (changes requested):

roles/debian-hardening/vars/main.yml
lucbeaufils: In ANSSI doc: Disabling mce will force a kernel panic at uncorrect error reported by Machine Check instead of a send of SIGBUS which can ba exploiting by malware like Rowhammer

deribaucourt (approved):

roles/debian-hardening/vars/main.yml
```diff
@@ -12,6 +12,8 @@ kernel_params:
  - slab_common.usercopy_fallback=N
  - iommu=pt
  - security=yama
+ - mce=off
```
dupremathieu, line 15: I was wrong in my previous comment. mce=0 does not disable MCE but sets the tolerance level to 0, in a result, all MCE events will force Linux to panic.
So mce=0 was correct.

dupremathieu (changes requested):

dupremathieu (approved):

deribaucourt (approved):

PR #177, opened 2023-03-22
cukinia_tests: use template for apt.conf

deribaucourt:
sources.list can now be configured to use an apt proxy instead of official repositories.

SFL: #9314
Change-Id: Ibc5ef5f90cce78d27f35d515611ee9c6dc3fcf9f

### DIFF
--- a/playbooks/test_deploy_cukinia_tests.yaml
+++ b/playbooks/test_deploy_cukinia_tests.yaml
@@ -9,6 +9,12 @@
       copy:
         src: ../src/cukinia-tests/cukinia
         dest: /etc/
+    - name: Copy Cukinia's tests templates
+      template:
+        src: ../src/cukinia-tests/cukinia/{{ item.src }}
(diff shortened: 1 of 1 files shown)


playbooks/test_deploy_cukinia_tests.yaml
```diff
@@ -9,6 +9,12 @@
       copy:
         src: ../src/cukinia-tests/cukinia
         dest: /etc/
+    - name: Copy Cukinia's tests templates
+      template:
```
dupremathieu, line 13: apt.conf is not a template. I suggest using copy instead of template.

dupremathieu (changes requested):

playbooks/test_deploy_cukinia_tests.yaml
dupremathieu: Ok I was wrong, it is a template. Maybe we can rename the file to add the .j2 extension.

dupremathieu (approved):

PR #186, opened 2023-04-19
Ansi bp 028 r39

eroussy:
This PR add security tags for sudo and correct the bugs caused by this


dupremathieu (commented):
If Ansible has to use the ansible user, it must be created by FAI, correct?

eroussy:
> If Ansible has to use the ansible user, it must be created by FAI, correct?

Yes, and it is already

dupremathieu (approved):

PR #200, opened 2023-05-16
roles/debian-hardening: disable ipv6 for all machines

eroussy:
Previously, ipv6 was disable in a hypervisor file. The observer still had ipv6 enabled.


insatomcat:
I think the change should be done in the debian playbook, rather than in the hardening playbook.
Please take a look at #201.
Converting this PR as a draft in the meantime. Thanks.

PR #225, opened 2023-05-31
Adding the group virtu before the user virtu

laurent-morin:
When deploying cluster_setup_prerequisdebian.yaml on a fresh Debian install, I had an error at the virtu user creation where the virtu group where missing.

### DIFF
--- a/roles/debian/tasks/main.yml
+++ b/roles/debian/tasks/main.yml
@@ -260,6 +260,11 @@
     name: systemd-journald
   when: journaldconf.changed
 
+- name: Create group virtu before the user
+  group :
+    name: virtu
+    state: present
+
 - name: Adding user virtu to groups libvirt and haclient
   user: name=virtu
     group=virtu


dupremathieu:
SEAPATH cannot be used on a regular Debian. You have to generate your Debian ISO using https://github.com/seapath/build_debian_iso. The virtu user is already created if you use this ISO.

laurent-morin:
We are using the iso build with build_debian_iso. In our configuration we set our admin user instead of virtu.
Does the virtu user is needed maybe the admin user could be a variable in the inventory.

dupremathieu (changes requested):
Could you please sign your commit and add a commit message body as described in our contributing guide: https://github.com/seapath/.github/blob/main/CONTRIBUTING.md?

insatomcat:
Hello,
Right now the build_debian_iso logic is linked to the ansible logic. We create an admin user in the iso (virtu) and deal with its privileges in multiple ansible playbooks (debian role, hardening role, cluster_setup_keys playbook, etc.).
This user has multiple uses (which will need to be splitted in the future):
1) "dev" or "rescue" user that that connect to the servers with ssh and sudo as root
2) used by libvirt to live-migrate the guests from one host to another (hence the cluster_setup_keys playbook)

What is your need exactly ?
If it is to change "virtu" to "admin", then you will have to propose a "variabilization" of the virtu username (accross all roles that uses this "virtu" username).

Once again, we also need to think that in the future, this virtu/admin user will not be the one used for live-migration.

The best approach would probably be to work on one unique PR : 
- splitting the use of virtu (into one "rescue" account and one "live-migration" account),
- variabilize both those usernames.

insatomcat:
will solve in #255

PR #236, opened 2023-06-06
debian prereq: remove all FAI grub defaults

insatomcat:
Following https://github.com/seapath/build_debian_iso/pull/57 FAI now has multiple default files for grub
Ansible needs to remove all of them, hence the use of a wildcard pattern (we remove all files in /etc/default/grub.d starting with 2 digits and ending in .cfg)


insatomcat:
I'll add the console parameters to the debian roles.

dupremathieu (approved):

PR #241, opened 2023-06-09
CI: reduce number of reboots

insatomcat:


playbooks/ci_restore_snapshot.yaml
```diff
@@ -7,15 +7,24 @@
   hosts: cluster_machines
   become: true
   tasks:
+    - name: "grub conf"
```
eroussy, line 10: this task is already present in the debian roles, is it really useful to do it here ?

insatomcat: yes, the idea here is that we are rollbacking the machines to zero. The LVM rollback only applies to vg1/root, and the grub configuration is in /boot, so we need this task to rollback the grub config.

eroussy (approved):

PR #245, opened 2023-06-13
Some bugs

insatomcat:
- 01-init.network file was not properly removed
- systemd-resolved was disabled by hardening playbook (it shouldn't)
- systemd-resolved was not bound on a specific address (because af LLMNR)
- vim was not the default editor (vi was, which created some problem with the default vim configuration)
- we also bump the cukinia-tests submodule to take into account the systemd-resolved changes


dupremathieu (changes requested):
Can you update the commit message to add a brief description of the fixes?

dupremathieu (approved):

PR #246, opened 2023-06-14
Solved timemaster long delay at start

insatomcat:
due to the dependancy on sys-devices-virtual-net-ptp.device


templates/timemaster.service.j2
```diff
@@ -1,3 +1,6 @@
 [Unit]
+After=
+After=chronyd.service ntpd.service ntpdate.service sntp.service
```
dupremathieu, line 3: I know the vanilla service add After chronyd.service ntpd.service but there is also a Conflicts chronyd.service ntpd.service, so you can remove chronyd.service and ntpd.service from After.

dupremathieu (approved):

templates/timemaster.service.j2
insatomcat: The default is:
```
[Unit]
Description=Synchronize system clock to NTP and PTP time sources
Documentation=man:timemaster
After=chronyd.service ntpd.service ntpdate.service sntp.service sys-devices-virtual-net-ptp.device
Wants=sys-devices-virtual-net-ptp.device
Conflicts=chronyd.service ntpd.service phc2sys.service ptp4l.service```

I just tried to override After and Wants, since they are the only ones referencing sys-devices-virtual-net-ptp.device

Does not seem to be enough though...

PR #254, opened 2023-07-03
ntp: allow more than 2 NTP servers

insatomcat:
Replacing ntp_primary_server and ntp_secondary_server variables with ntp_servers which is an array (not limited to 2 items). 
The first item will be the preferred server if no ptp interface is defined. 
For now the change is retro-compatible.


insatomcat:
good point!

dupremathieu (approved):

PR #256, opened 2023-07-05
restore.sh: remove hardcoded "virtu" user

insatomcat:
We are now using the "livemigration_user" variable.

### DIFF
--- a/roles/debian/tasks/main.yml
+++ b/roles/debian/tasks/main.yml
@@ -51,6 +51,16 @@
     dest: /usr/local/bin/
     rsync_opts:
       - "--chown=root:root"
+      - "--exclude=*.j2"
+- name: Copy backup-restore templates
+  template:
+    src: ../src/debian/backup-restore/{{ item.src }}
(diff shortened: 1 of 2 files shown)


roles/debian/tasks/main.yml
```diff
@@ -51,6 +51,14 @@
     dest: /usr/local/bin/
     rsync_opts:
       - "--chown=root:root"
+      - "--exclude=*.j2"
+- name: Copy backup-restore templates
+  template:
+    src: ../src/debian/backup-restore/{{ item.src }}
+    dest: /usr/local/bin/{{ item.dest }}
+  with_items:
+      - { src: 'restore_vm.sh.j2',
```
dupremathieu, line 60: This syntax is not very "readable".
I prefered:
```yaml
  with_items:
    - src:
       restore_vm.sh.j2
      dest:
        restore_vm.sh
```

dupremathieu (approved):

dupremathieu (changes requested):
See comment

PR #257, opened 2023-07-06
Allow ceph osd to only use part of disk instead of all disk, to allow for lv snapshot

tpeypelu:
By default, ceph osd will use an entire disk (ceph_osd_disk). This does not allow to use LVM snapshots for ceph, which can be a problem when doing big upgrades (like debian os major version): no rollback is possible.
This commit will allow the user to only use a partition of a disk, create the ceph VG on it, and then use only part of this VG for the ceph LV, leaving some place for a future snapshot.
This change is retrocompatible, the user will need to specifically set the lvm_volumes object in the inventory to use custom partition/VG/LV


inventories/seapath_cluster_definition_example.yml
```diff
@@ -127,6 +121,18 @@ all:
                 node1:
                 node2:
                 node3:
+            vars:
+                # Ceph settings
+                ceph_osd_disk: /dev/sdb # CEPH OSD disk
+#                lvm_volumes:
```
dupremathieu, line 127: Is it possible to prefix lvm_volumes by ceph_?

inventories/seapath_cluster_definition_example.yml
```diff
@@ -127,6 +121,18 @@ all:
...
+#                    - data: lv_ceph # Name of the logical volume to use for the CEPH OSD volume
+#                      data_vg: vg_ceph # Name of the volume group to use for the CEPH OSD volume
+#                      data_size: 5G # Size of the logical volume, default in megabytes, possible values: [0-9]+[bBsSkKmMgGtTpPeE]
+#                      device: "{{ ceph_osd_disk }}"
+#                      device_number: 1 # Number of the partition in the ceph_osd_disk
+#                      device_size: 7GiB # Size of the partition in the ceph_osd_disk, default in megabytes, possible units: B, KB, KiB, MB, MiB, GB, GiB, TB, TiB
+                devices:
```
dupremathieu, line 134: Same remark for devices. I prefer to prefix it with _ceph.

dupremathieu (changes requested):

inventories/seapath_cluster_definition_example.yml
dupremathieu: I was not aware it was a ceph-ansible variable. So, it is ok in that case.

inventories/seapath_cluster_definition_example.yml
dupremathieu: Ditto

dupremathieu (approved):

PR #278, opened 2023-08-07
ceph: plan for the case ceph VG/LG already exist

insatomcat:
This change has to be merged ASAP for the CI to be functionnal

It might be a good target to assume the ceph VG and LG have already been created (for example for a softraid situation). In that case the ceph playbook must deal with multiple possibilities:
- Is it a new ceph installation or not ?
- if it's a new installation, do we already have the ceph VG and LV, in which case we zap them and move to the installation
- if the ceph VG and LV don't exist, are we really in a "customized lvm situation"
- if we are, then we create what the inventory requires, then move to the installation
- if we are not, then we use the whole "ceph_osd_disk" disk and move to the installation

This commit adds the "existing ceph VG/LG case".


insatomcat:
included in #277

PR #281, opened 2023-08-16
switch from hddtemp to smartmontools

insatomcat:
hddtemp is used to monitor the hard disks temperature. This package is dormant and is not available on debian12. It is recommended to use smartmontools instead.
This commit creates the snmp endpoint to retrieve the hard drive temperatures using Smartmontools, and remove the hddtemp deployment code from the playbooks.


src/debian/snmp_tempdisk.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
```
eroussy, line 1: This file is never copied on the system.
A line must be added in the playbook to copy it in the hypervisor.

eroussy (changes requested):

src/debian/snmp_tempdisk.sh
insatomcat: the file is copied by this task/line : https://github.com/seapath/ansible/blob/2311d8fcbafc81fb40c7fbbd34c9712d179dd919/roles/debian/tasks/main.yml#L228C13-L228C13

eroussy (approved):

PR #291, opened 2023-08-23
library/cluster_vm: set crm_config_cmd default value to None

dupremathieu:
The default crm_config_cmd value has to be None and not [].


dupremathieu:
Fix in #292.

PR #302, opened 2023-09-16
Merge all waiting changes

insatomcat:
Dockerfile: move from Ubuntu to Debian
For the sake of consistency, it seems logical to use Debian in this container rather than Ubuntu.
-----------------------
lvm_snapshot_rebooter: manual polling
With this commit we only activate the necessary LV (the ones that actually have a pending rollback) and we poll them manually so that we know exactly when the rollback is completed.
-----------------------
ha playbook: make extra cmd run by one of the cluster node
This task needs to be run on a cluster node. If it runs on a standalone node, it won't work.
-----------------------
80-wired file: revert commit 040f3bf
-----------------------
inventory: correct typos
-----------------------
deploy_vm: check for vm presence
-----------------------
vm inventory: improve crm location examples
-----------------------
inventory: main_disk is not useful anymore
it was for hddtemp only
-----------------------
VM inventory: allow customization of qcow2 and xml files
-----------------------
deploy VM: change tmp folder
In case the qcow2 file is to big for /tmp
-----------------------
pacemaker systemd: allow the shutdown timeout to be configurable
pacemaker shutdown timeout is 30min by default. This is long because if there is a problem (networking, stuck vm, etc.) preventing shutdown, the server will be stuck for two times the shutdown timeout (60min).
Seapath used to reduce this timeout to 2min, but this can be too short if there are many Guests to migrate before shutdown.
With this commit the timeout is 2min by default, but can be changed with an inventory variable.
-----------------------
deploy_vm: remove linefeed in xml data
This is to prevent the resulting log to be "multi-lines" which will be impossible to parse & export for syslog-ng
-----------------------
inventory: make logstash_server_ip optional
This variable is only useful when deploying and using metricbeat/filebeat
-----------------------
libvirtd: force daemon facility when exporting libvirt logs
libvirtd has a bug where it does not correctly set its syslog facility. This will make the syslog recipient server choke on the log.
It will be fixed in debian12 (https://github.com/libvirt/libvirt/commit/017fa1e097769455c0cd8929760cf9ff52a556ec)
In the meantime, we make syslog-ng force the facility to daemon (3) before exporting the logs.


.github/workflows/ci-debian.yml
```diff
@@ -19,7 +19,7 @@ jobs:
 
       - name: Initialize sources
         run: mkdir ${{ env.WORK_DIR }}; cd ${{ env.WORK_DIR }};
-             git clone -q --depth 1 -b main https://github.com/seapath/ci ci;
+             git clone -q --depth 1 -b paramvm https://github.com/seapath/ci ci;
```
dupremathieu, line 22: I do not understand, you just change the branch.

.gitmodules
```diff
@@ -10,7 +10,7 @@
 [submodule "src/debian/vm_manager"]
 	path = src/debian/vm_manager
 	url = https://github.com/seapath/vm_manager.git
-        branch = main
+        branch = debian12
```
dupremathieu, line 13: I approved the vm_manager PR so no need to change the branch.

.gitmodules
```diff
@@ -14,7 +14,7 @@
 [submodule "src/cukinia-tests"]
 	path = src/cukinia-tests
 	url = https://github.com/seapath/cukinia-tests.git
-        branch = main
+        branch = evol
```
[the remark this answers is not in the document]
insatomcat: true, I intend to change this after we merge the cukinia-tests PR

.github/workflows/ci-debian.yml
insatomcat: I had to, as look as https://github.com/seapath/ci/pull/46 was not merged.
It now is so I will change this back to main.

prepare.sh
```diff
@@ -35,7 +35,7 @@ if ! command -v ansible-galaxy &>/dev/null ; then
 fi
 
 echo "Test python3 netaddr module is installed"
-if ! cat << EOF | python3 &>/dev/null
+if ! cat << EOF | /env/bin/python3 &>/dev/null
```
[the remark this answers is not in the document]
insatomcat: I forced /env/bin/python3 as a default in the docker image and removed this change from prepare.sh

insatomcat:
> commit [f761d61](https://github.com/seapath/ansible/commit/f761d61df70c0fe7f6a6a4b235e54b146dd79ac7) the title is to long

shortened as requested

dupremathieu (approved):

PR #305, opened 2023-10-01
waiting changes

insatomcat:
**example inventory: cleanup + add SR-IOV example**

----
**add support for vlan on br0**

Seapath configures the main network interface of the host in a bridge called br0. This is to allow a Guest to easily bridge on this interface, to give that guest easy remote access. 
In some network topologies, this main host network interface is a trunk, and the host need to support vlan tagging to get access to the network. 
This commit makes this topology supported, throught the use of the "br0vlan" configuration variable (in which the user simply has to set the vlanid number).

Then we have to provide a way to configure the guests access to br0. 
This is now possible with the "guest_on_br0" variable in which we define the tap interfaces for the guests (port access or trunk are possible). The user will then only have to configure its guests (livbirt xml) to use those taps, exactly like what we do with OVS, like this:
 ```
    <interface type="ethernet">
      <target dev="GUESTtap" managed="no"/>
      <model type="virtio" />
      <address type="pci" domain="0x0000" bus="0x00" slot="0x01" function="0x0" />
    </interface>
```

----
**add support for HSR for cluster network**

Debian 12 comes with a backported kernel including the HSR module, allowing to use HSR for cluster networking instead of RSTP.
This commit make it possible to choose between the two.

----
**systemd ordering: add chrony-wait to time-sync.target**

Many services depends on the time-sync.target (ceph, pacemaker...). However the time-sync.target is not linked to anything right now. This commit add the chrony-wait service, inspired from what's coming in chrony with debian12, to the time-sync.target.
This means ceph and pacemaker will wait for chrony to be synced.
Also we make pacemaker depend on libvirtd, just to be sure the system is able to run guests before we launch pacemaker.

----
**solves bug in disabling systemd-timesyncd**

----
**example inventory: network reboot params documentation**

### DIFF
--- a/inventories/seapath_cluster_definition_example.yml
+++ b/inventories/seapath_cluster_definition_example.yml
@@ -31,6 +31,18 @@ all:
                             cluster_ip_addr: "192.168.55.1"
                             br_rstp_priority: 12288
                             brBRIDGE1_ext: eno1234 #physical nic to use with BRIDGE1 (see the ovs topology inventory)
+                            custom_network:
+                              - 01-enp24s0_sriov:
+                                - Match:
+                                  - Name: "enp24s0*"
+                                - SR-IOV:
+                                  - VirtualFunction: 0
+                                  - Trust: 1
+                                  - LinkState: on
+                                - SR-IOV:
+                                  - VirtualFunction: 1
+                                  - Trust: 1
+                                  - LinkState: on
 
                         node2:
                             # ansbile variables
@@ -246,11 +258,11 @@ all:
         # Optional list variable to change apt repositories. All repositories will be overrides.
         # If defined as empty list [], all apt repositories will be removed
         apt_repo:
-            - http://ftp.fr.debian.org/debian bullseye main contrib non-free
-            - http://security.debian.org/debian-security bullseye-security main contrib non-free
-            - http://ftp.fr.debian.org/debian bullseye-backports main contrib non-free
-            - https://download.docker.com/linux/debian bullseye stable
-            - https://artifacts.elastic.co/packages/8.x/apt stable main
+          - http://ftp.fr.debian.org/debian bullseye main contrib non-free
+          - http://security.debian.org/debian-security bullseye-security main contrib non-free
+          - http://ftp.fr.debian.org/debian bullseye-backports main contrib non-free
+          - https://download.docker.com/linux/debian bullseye stable
+          - https://artifacts.elastic.co/packages/8.x/apt stable main
 
         # Default user with admin privileges, and password hash
         admin_user: virtu
--- a/inventories/seapath_cluster_definition_example.yml
+++ b/inventories/seapath_cluster_definition_example.yml
@@ -294,3 +294,4 @@ all:
           GUEST4: []        # no filtering, including untagged trafic
 
         #cluster_protocol: "HSR" # RSTP or HSR, default is RSTP
+        chrony_wait_timeout_sec: 180 #time in seconds the boot sequence will wait for chrony to sync
--- a/inventories/seapath_cluster_definition_example.yml
+++ b/inventories/seapath_cluster_definition_example.yml
@@ -210,7 +210,8 @@ all:
         # Main network configuration
         gateway_addr: 10.0.0.1
         dns_servers: "8.8.8.8 8.8.4.4"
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

File: src/debian/pacemaker_override.conf.j2 [hunk 1]
```diff
@@ -1,7 +1,6 @@
 [Unit]
-After=rbdmap.service
-After=libvirt-guests.service
-After=ceph.target
```
dupremathieu, line 4: We do not need to want for ceph and rbdmap?

inventories/seapath_cluster_definition_example.yml
```diff
@@ -246,11 +259,11 @@ all:
         # Optional list variable to change apt repositories. All repositories will be overrides.
         # If defined as empty list [], all apt repositories will be removed
         apt_repo:
-            - http://ftp.fr.debian.org/debian bullseye main contrib non-free
-            - http://security.debian.org/debian-security bullseye-security main contrib non-free
-            - http://ftp.fr.debian.org/debian bullseye-backports main contrib non-free
-            - https://download.docker.com/linux/debian bullseye stable
-            - https://artifacts.elastic.co/packages/8.x/apt stable main
+          - http://ftp.fr.debian.org/debian bullseye main contrib non-free
```
dupremathieu, line 265: This change is not relative with this commit.

inventories/seapath_cluster_definition_example.yml
```diff
@@ -31,6 +32,18 @@ all:
                             cluster_ip_addr: "192.168.55.1"
                             br_rstp_priority: 12288
                             brBRIDGE1_ext: eno1234 #physical nic to use with BRIDGE1 (see the ovs topology inventory)
+                            custom_network:
+                              - 01-enp24s0_sriov:
```
dupremathieu, line 37: This change is not relative with this commit.

dupremathieu (changes requested):
For commit titles, if you want to respect our contribution guide, commit's title must be in imperative mood.
See: https://github.com/seapath/.github/blob/main/CONTRIBUTING.md#git-workflow-guidelines
For instance: add support vlan on br0

I am ok with adding HSR support, but for the record, without hardware assistant it adds a lot of additional processing. For me, the best is to use RSTP with mstpd (instead of OVS).

For the bridge vlan, I found the changes are heavy. Why not remove the regular Linux bridge to use an OVS bridge instead to re-use all the works done with the VLAN?

src/debian/pacemaker_override.conf.j2
insatomcat: libvirt-guests.service is disabled right now.
ceph is already a dependency of rbdmap.service
as for rbdmap, it already is a dependency of libvirtd (via remote-fs.target), so I can replace it with a direct dependency of pacemaker to libvirtd.

inventories/seapath_cluster_definition_example.yml
insatomcat: indeed, I will split the commit

insatomcat:
> For commit titles, if you want to respect our contribution guide, commit's title must be in imperative mood. See: https://github.com/seapath/.github/blob/main/CONTRIBUTING.md#git-workflow-guidelines For instance: add support vlan on br0

OK, I rewrote the titles. 

> I am ok with adding HSR support, but for the record, without hardware assistant it adds a lot of additional processing. For me, the best is to use RSTP with mstpd (instead of OVS).

You are right, but I'm still investigating RSTP potential impact on corosync quorum. What would happen if rstp took longer than a second to converge? Corosync would go into fencing mode...
HSR is merely a try at plan B.

Why do you think mstpd is better than ovs in that case?

> For the bridge vlan, I found the changes are heavy. Why not remove the regular Linux bridge to use an OVS bridge instead to re-use all the works done with the VLAN?

br0 will be the "main" interface of the host, so I think that it must have the least dependencies possible. I would not like it to lose my ssh connection to the host if ovs has a pb... hence the choice to use networkd.

dupremathieu (changes requested):
example inventory: cleanup + add ST-IOV example  -> example inventory: cleanup + add SR-IOV example

insatomcat:
> example inventory: cleanup + add ST-IOV example -> example inventory: cleanup + add SR-IOV example

done

PR #313, opened 2023-10-31
snmp: adds lvs and smartctl status

insatomcat:


src/debian/snmp_smartctl.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
```
dupremathieu, line 1: `#!/bin/bash`

src/debian/snmp_smartctl.sh
```diff
@@ -0,0 +1,9 @@
...
+
+/usr/sbin/smartctl --scan | grep -E "^/dev/sd" | awk '{ print $1 }' | while read i; do
```
dupremathieu, line 3: Also look for nvme device: `grep -E '^/dev/(sd|nvme)'`.

dupremathieu (changes requested):

src/debian/snmp_smartctl.sh
insatomcat: ok

dupremathieu (approved):

PR #317, opened 2023-11-12
Adds HTTP API to vm_manager

insatomcat:
This commit will add way to control vm-manager with an HTTP request on any node. 

1) A docker container running apache-php will listen on port 8080
   It will wait for the command to run and for a security code.
   It will not check the security code itself
   It will check the provided command to run starts with "vm-mgr" and only contains needed characters (to prevent command injection attacks)
   If all is correct, it will send the code and the requested command to a FIFO pipe (see step 2 now)
   Once sent, it will read the same FIFO pipe to wait for the return of the command
   It will then output the return of the command

2) A systemd service will read the fifo pipe in a loop
   at every loop, it will check the fifo pipe exists and has the correct permissions
   it will read the pipe and wait for a command
   Once the code+command arrive, it will check the provided security code is the same as the content of file /vmmgrapi.secret (this file has to be manually created by root on all nodes + chmod 600)
   it will then run the command, with a 10s timeout, and write the output of the command to the pipe

Both those services must be managed with pacemaker (they are linked systemd units).
In order to respect the bindind socket rule (that any service must bind on a specific IP address), Pacemaker must manage an IpAddr2 resource, colocated with the docker systemd service. The docker container will only listen to the "moving" ip address.
Examples are provided in the default inventory.

### DIFF
--- a/inventories/seapath_cluster_definition_example.yml
+++ b/inventories/seapath_cluster_definition_example.yml
@@ -279,12 +279,14 @@ all:
         livemigration_user: livemigration
 
         #pacemaker_shutdown_timeout: "2min"
+        #enable_vmmgr_http_api: false
+        #cluster_ip: 10.10.11.4 # IP the vm-mgr http api will bind on
         extra_crm_cmd_to_run:
           # for ntp status monitoting, be sure host_ip is the ip address of the server and not its hostname
-          - primitive ntpstatus_test ocf:seapath:ntpstatus params host_ip=10.10.10.10 multiplier=500 op monitor timeout=10 interval=10
-          - primitive ptpstatus_test ocf:seapath:ptpstatus op monitor timeout=10 interval=10 op_params multiplier=1000
-          - clone cl_ntpstatus_test ntpstatus_test meta target-role=Started
-          - clone cl_ptpstatus_test ptpstatus_test meta target-role=Started
+          - "primitive ntpstatus_test ocf:seapath:ntpstatus params host_ip=10.10.10.10 multiplier=500 op monitor timeout=10 interval=10"
+          - "primitive ptpstatus_test ocf:seapath:ptpstatus op monitor timeout=10 interval=10 op_params multiplier=1000"
+          - "clone cl_ntpstatus_test ntpstatus_test meta target-role=Started"
+          - "clone cl_ptpstatus_test ptpstatus_test meta target-role=Started"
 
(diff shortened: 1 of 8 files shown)


dupremathieu (changes requested):
We cannot add an HTTP interface for vm-manager in that way.
It is a very hacky solution and huge security hole.
* it relies on CLI, which is not designed for this case
* it uses unencrypted connection for a critical service
* it bypasses all SEAPATH authentication mechanisms
* it does not log the authentication
* there are no trust mechanisms

The best will be to modify vm_manager to add a REST API.
For instance:
1. **List VMs**
   - **Endpoint:** `GET /vms`
   - **Description:** Lists all VMs.

2. **VM Operations (Start, Stop, Remove, Enable, Disable, Status)**
   - **Endpoint:** `POST /vms/{vm_name}/{action}`
   - **Description:** Perform actions like start, stop, remove, etc., on a VM.
   - **Path Parameters:**
     - `vm_name`: The name of the VM.
     - `action`: The action to perform (start, stop, remove, enable, disable, status).

3. **Create VM**
   - **Endpoint:** `POST /vms/create`
   - **Body Parameters:**
     - `name`: VM name.
     - `xml`: VM libvirt XML data.
     - Additional parameters as needed (e.g., `image`, `disable`, `metadata`, etc.).

4. **Clone VM**
   - **Endpoint:** `POST /vms/clone`
   - **Body Parameters:**
     - `name`: Source VM name.
     - `dst_name`: Destination VM name.
     - Additional parameters similar to create VM.

5. **Snapshot Management (Create, Remove, List, Purge, Rollback)**
   - **Endpoint:** `POST /vms/{vm_name}/snapshots/{action}`
   - **Description:** Manage VM snapshots.
   - **Path Parameters:**
     - `vm_name`: The name of the VM.
     - `action`: The snapshot action (create, remove, list, purge, rollback).
   - **Body Parameters:**
     - Specific to action, like `snap_name`, `date`, `number`, etc.

6. **Metadata Operations (List, Get, Set)**
   - **Endpoint:** `POST /vms/{vm_name}/metadata/{action}`
   - **Description:** Operations on VM metadata.
   - **Path Parameters:**
     - `vm_name`: The name of the VM.
     - `action`: The metadata action (list, get, set).
   - **Body Parameters:**
     - For `get` and `set`, parameters like `metadata_name`, `metadata_value`.

7. **Colocation Management**
   - **Endpoint:** `POST /vms/{vm_name}/colocation`
   - **Description:** Add a colocation constraint.
   - **Path Parameters:**
     - `vm_name`: The name of the VM.
   - **Body Parameters:**
     - `resources`: List of VMs or resources.
     - `strong`: Boolean indicating if it's a strong colocation constraint.

PR #320, opened 2023-11-15
adds vm_manager flask REST api

insatomcat:
adds vm_manager flask REST api

vm_manager provides a basic flask REST api.
This commit configures SEAPATH to securely expose this API (can be enabled/disabled in the inventory).
Since the API is python based, we use nginx and gunicorn (which will need to be installed on the hosts, along with python3-flask and python3-flaskext.wtf used for CSRF protection)..

- All node will have a gunicorn systemd socket running
- This socket will belong to www-data with chmod 600 (/run/gunicorn.sock) : only the www-data user (nginx) can use the API. Not even a local user can use it.
- This socket is bound to the gunicorn service, exposing the vm_mgr api. This service will only be launched when needed.

- nginx will be used to expose the gunicorn socket
- nginx will be managed by pacemaker as a systemd resource
- this resource will be colocated with a IPaddr2 resource, exposing a "cluster administration IP" (for this pacemaker will need the cap_net_admin added in its systemd unit file)
- nginx will only listen on this cluster IP

- nginx will use HTTPS for secure communication. The seapath user can use its own certificate (configurable in the inventory) but by default, autosigned certificated will be created and used.
- nginx can use an "allowed ip" access list, configurable in the inventory
- nginx will authenticate the connection with an htpasswd file (must be configured by the seapath user). The local path
 of this file is configurable in the inventory.

### DIFF
--- a/src/debian/vmmgrapi/gunicorn.service
+++ b/src/debian/vmmgrapi/gunicorn.service
@@ -1,0 +1,23 @@
+[Unit]
+Description=gunicorn daemon
+Requires=gunicorn.socket
+After=network.target
+
+[Service]
+Type=notify
+# the specific user that our service will run as
+User=root
+Group=root
+# another option for an even more restricted service is
+# DynamicUser=yes
+# see http://0pointer.net/blog/dynamic-users-with-systemd.html
+RuntimeDirectory=gunicorn
+WorkingDirectory=/var/local/vmmgrapi
+ExecStart=/usr/bin/gunicorn wsgi:app
+ExecReload=/bin/kill -s HUP $MAINPID
+KillMode=mixed
+TimeoutStopSec=5
+PrivateTmp=true
+
+[Install]
+WantedBy=multi-user.target
--- a/src/debian/vmmgrapi/nginx.conf.j2
+++ b/src/debian/vmmgrapi/nginx.conf.j2
@@ -1,0 +1,44 @@
+worker_processes 1;
+user www-data www-data;
+error_log  /var/log/nginx/error.log warn;
+pid /var/run/nginx.pid;
+events {
+  worker_connections 1024;
+}
+http {
+  access_log /var/log/nginx/access.log combined;
+  upstream app_server {
+    server unix:/run/gunicorn.sock fail_timeout=0;
+  }
+  server {
+    listen {{ admin_cluster_ip }}:{{ vmmgr_http_port }} ssl;
+    ssl_certificate {{ vmmgrapi_certs_dir }}/seapath.crt;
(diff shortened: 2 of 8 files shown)


src/debian/vmmgrapi/gunicorn.service
```diff
@@ -0,0 +1,23 @@
+[Unit]
+Description=gunicorn daemon
+Requires=gunicorn.socket
+After=network.target
+
+[Service]
+Type=notify
+# the specific user that our service will run as
+User=root
```
dupremathieu, line 9: Is it possible to not launch gunicorn with another user than root?

src/debian/vmmgrapi/nginx.conf.j2
```diff
@@ -0,0 +1,44 @@
...
+events {
+  worker_connections 1024;
+}
+http {
+  access_log /var/log/nginx/access.log combined;
+  upstream app_server {
+    server unix:/run/gunicorn.sock fail_timeout=0;
+  }
+  server {
+    listen {{ admin_cluster_ip }}:{{ vmmgr_http_port }} ssl;
```
dupremathieu, line 14: You can use the Mozilla SSL configurator to harden the TLS connection: https://ssl-config.mozilla.org/#server=nginx&version=1.17.7&config=modern&openssl=1.1.1k&guideline=5.7

src/debian/vmmgrapi/nginx.conf.j2
```diff
@@ -0,0 +1,44 @@
...
+
+    {% if vmmgr_http_api_acl is defined -%}
+    {{ vmmgr_http_api_acl | indent( width=4, first=False) }}
+    {%- endif %}
+
+    server_name seapath;
+    keepalive_timeout 5;
+    location / {
+      {% if vmmgr_http_local_auth_file is defined -%}
+      auth_basic           "RESTRICTED Area";
```
dupremathieu, line 26: Depending on the use case, there is also the auth_pam module that can be useful. With this module, you can use the system user and the LDAP user. It requires installing the package nginx-extras on Debian.

dupremathieu (changes requested):
Only running the gunicorn service in root is important. The other is nice to have.

dupremathieu (approved):

src/debian/vmmgrapi/gunicorn.service
insatomcat: for now, vm-mgr requires the root user (needs to run some crm and rbd commands)

src/debian/vmmgrapi/nginx.conf.j2
insatomcat: might be a nice future addition yes

PR #324, opened 2023-11-23
virt-df.sh: adds support for UFS filesystem (for BSD guests)

insatomcat:


src/debian/virt-df.sh
```diff
@@ -36,6 +36,10 @@ do
   do
     /usr/bin/mount -o ro,noload $part /t 2>/dev/null
     returncode=$?
+    if [ $returncode -ne 0 ]; then
```
dupremathieu, line 39: I preferred trying to detect the fstype.
```sh
if [ $(lsblk -o FSTYPE --noheadings $part) == ufs ] ; then
  /usr/bin/mount -o ro,ufstype=ufs2 -t ufs $part /t 2>/dev/null
  returncode=$?
else
  /usr/bin/mount -o ro,noload $part /t 2>/dev/null
  returncode=$?
fi
```

dupremathieu (changes requested):

dupremathieu (approved):

src/debian/virt-df.sh
insatomcat: neat idea, but it does not seem to work on rbd devices

PR #333, opened 2023-12-11
openflow: provides a way to override the whole flow tables

insatomcat:
The setup_ovs module allows the seapath user to use openflow rules to enable/disable ipv6 between guests, and to do antispoofing (based on mac/ip addresses of guests). There is currently no way to do more than that.

This commits allows the user to override the openflow table with custom rules, using the "flows_override" inventory variable.


insatomcat:
standing by merge of PR https://github.com/seapath/python3-setup-ovs/pull/12

dupremathieu (approved):

PR #335, opened 2023-12-20
network_vars: remove vlan filtering on physical port

insatomcat:
In case of VLAN filtering, we let the physical port accept ALL vlan ids by defau
lts (1-4094)

### DIFF
--- a/vars/network_vars.yml
+++ b/vars/network_vars.yml
@@ -38,9 +38,7 @@ br0vlan_systemd_networkd_network:
         - Name: "{{ network_interface }}"
     - Network: "{{ [ {'Bridge': 'br0'} ] + extra_network_config | default([]) }}"
     - BridgeVLAN:
-        - VLAN: 159
-        - PVID: 4094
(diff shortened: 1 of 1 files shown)


vars/network_vars.yml
```diff
@@ -38,7 +38,7 @@ br0vlan_systemd_networkd_network:
         - Name: "{{ network_interface }}"
     - Network: "{{ [ {'Bridge': 'br0'} ] + extra_network_config | default([]) }}"
     - BridgeVLAN:
-        - VLAN: 159
+        - VLAN: "{{ br0vlan | default('') }}"
         - PVID: 4094
```
dupremathieu, line 42: There are the same issue PVID and EgressUntagged.

dupremathieu (changes requested):

vars/network_vars.yml
insatomcat: I'm going to remove the whole section

dupremathieu (approved):

PR #341, opened 2024-01-10
Update example inventory

eroussy:


examples/inventories/advanced_inventory_example.yaml
```diff
@@ -227,3 +227,12 @@ all:
     # The vlan ID of the PTP. Remove the variable if the ptp frames
     # are not in a VLAN
     ptp_vlanid: 100
+    # Set to true to create a livemigration user used to migrate VM
+    # from one host to another (default is false)
+    # When no livemigration user is defined, root user is used. Doing
+    # so is considered a bad cybersecurity practice.
```
dupremathieu, line 243: For me, when no livemigration user is defined, the live migration is disabled. In any case, the live migration cannot be done using the root account because the root account is disabled.

dupremathieu (changes requested):

dupremathieu (approved):

PR #354, opened 2024-01-26
README.adoc: add needed dependencies

eroussy:
More packages are needed when building without cqfd.


dupremathieu (approved):
Could you also backpot this in the debian-main branche?

PR #357, opened 2024-01-31
bump setup_ovs + sysctl and networkd changes

insatomcat:
setup ovs: allow ofport_request option

bump python3-setup-ovs submodule to allow ofport_request option.
When setting up openflow rules based on incoming port, we need to know the openflow port number ("ofport") value, which is decided at port creation time.
Since the rules will use a fixed ofport, we need to fix the ofport for the ports we need openflow on.
OVS has the "ofport_request" option for this purpose.
This commit allows the seapath user to define the ofport_request value for a port in its OVS inventory, and it will be configured as requested at port creation.

----

sysctl: cleanup

This commits puts all sysctl files in a specific folder (src/debian/sysctl)

----

systcl: allows extra sysctl conf items for physical machines

This commits introduces the extra_sysctl_physical_machines variable, which content will be put in the /etc/sysctl.d/00-seapathextra_physicalmachines.conf on all physical machines.
This allows the seapath user to add extra sysctl conf items as needed (ip forwarding, ...).

----

network playbook: reload networkd instead of restart

Restarting networkd has more impact than reloading its configuration.
In case of an extra ip address managed by pacemaker by example, restart networkd will remove this extra address and have pacemaker recreate it (with an error).
Reloading networkd is lighter because it won't touch the interface if its configuration has not been modified.

----

snmp: better data curating in the smartctl exposition

We used to expose the whole "smartctl -a" output with snmp but this poses 2 problems:
- the disks were refered to with the standard (and non deterministic) sda, sdb naming scheme
- the smartctl -a output is often too big for a single snmp message

This commits reduces the data to information about the disk, result of self-test and attributes listing, but also adds the /dev/disk/by-path information to be certain of what disk the data refers to.


playbooks/cluster_setup_network.yaml
```diff
@@ -141,10 +141,9 @@
   vars:
     apply_config: "{{ apply_network_config | default(false) }}"
   tasks:
-    - name: Restart systemd-networkd
-      ansible.builtin.systemd:
-        name: systemd-networkd
-        state: restarted
+    - name: Reload systemd-networkd configuration
+      command:
```
dupremathieu, line 145: Why using command instead of systemd plugin?
The systemd supported the state "reloaded".
https://docs.ansible.com/ansible/2.9/modules/systemd_module.html

playbooks/cluster_setup_network.yaml
```diff
@@ -393,10 +392,9 @@
       with_items: "{{ guests_on_br0 | dict2items }}"
       when: guests_on_br0 is defined
       register: guest_br0_network
-    - name: Restart systemd-networkd
-      ansible.builtin.systemd:
-        name: systemd-networkd
-        state: restarted
+    - name: Reload systemd-networkd configuration
+      command:
```
dupremathieu, line 396: Ditto

dupremathieu (changes requested):

playbooks/cluster_setup_network.yaml
insatomcat: ok

playbooks/cluster_setup_network.yaml
insatomcat: ok

src/debian/snmp_smartctl.sh
```diff
@@ -1,9 +1,10 @@
 #!/bin/bash
 
 /usr/sbin/smartctl --scan | grep -E "^/dev/(sd|nvme)" | awk '{ print $1 }' | while read i; do
-  echo ------------------------------------------
   echo $i
-  echo ------------------------------------------
-  /usr/sbin/smartctl -a $i
+  j=$(echo $i | sed "s/\/dev\///")
```
dupremathieu, line 5: Move this in another commit.

dupremathieu (changes requested):

src/debian/snmp_smartctl.sh
insatomcat: oops, I forgot I had this in my working dir when pushing the above requested changes.
That's fixed.

dupremathieu (approved):

PR #358, opened 2024-01-31
Add vm example

eroussy:
Also refactor the example inventories so that they all be in one place.


dupremathieu (changes requested):
.cqfrd: update ansible-lint -> .cqfd: update ansible-lint

If you want to refactor the VM inventories doc, the best will be to have a dedicated adoc file, as it was done for Open vSwitch with OVS_configuration.adoc.

eroussy:
> If you want to refactor the VM inventories doc, the best will be to have a dedicated adoc file, as it was done for Open vSwitch with OVS_configuration.adoc.

You're right, but I will not be able to do that now. I only change the REAME to correct the path of the VM inventory.
I will open an issue for that.

dupremathieu (approved):

PR #365, opened 2024-02-06
Cpuset bug fix

eroussy:
Fix bug in vm template when the variable cpuset is not given

### DIFF
--- a/templates/vm/guest.xml.j2
+++ b/templates/vm/guest.xml.j2
@@ -8,6 +8,8 @@
   see examples/inventories/vms_inventory_example.yaml
 -->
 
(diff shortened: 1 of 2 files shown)


templates/vm/guest.xml.j2
```diff
@@ -57,19 +60,20 @@
         <vcpusched vcpus="{{ loop.index - 1 }}" scheduler="fifo" priority="{{ vm.rt_priority |default('1') }}" />
         {% endif %}
         {% endfor %}
+        {% endif %}
```
dupremathieu, line 63: Wrong indentation.

dupremathieu (changes requested):

dupremathieu (approved):

PR #391, opened 2024-02-19
Task order in cluster_setup_network.yaml playbook

Gabin-rte:
The task "configure guests' br0 interfaces" is executed after networkd configuration and can override custom network configurations from inventories. Reordering the task solves this issue.

### DIFF
--- a/playbooks/cluster_setup_network.yaml
+++ b/playbooks/cluster_setup_network.yaml
@@ -42,6 +42,41 @@
         owner: root
         group: systemd-network
 
+- name: configure guests br0 interfaces
+  hosts: cluster_machines
+  become: true
+  tasks:
+    - debug:
+        var: item.value
+      with_items: "{{ guests_on_br0 | dict2items }}"
+      when: guests_on_br0 is defined
+#      failed_when: 1 == 1
+    - name: create 00-GUESTtap.netdev
+      template:
+        src: ../templates/00-GUESTtap.netdev.j2
+        dest: /etc/systemd/network/00-{{ item.key }}tap.netdev
+        owner: "root"
+        group: "systemd-network"
+        mode: '0644'
+      with_items: "{{ guests_on_br0 | dict2items }}"
+      when: guests_on_br0 is defined
+      register: guest_br0_netdev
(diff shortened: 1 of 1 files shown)


playbooks/cluster_setup_network.yaml
```diff
@@ -42,6 +42,41 @@
...
+        group: "systemd-network"
+        mode: '0644'
+      with_items: "{{ guests_on_br0 | dict2items }}"
+      when: guests_on_br0 is defined
+      register: guest_br0_network
+    - name: Reload systemd-networkd configuration
+      ansible.builtin.service:
+        name: systemd-networkd
+        state: reloaded
+      when: guest_br0_netdev.changed or guest_br0_network.changed
```
insatomcat, line 78: The order of the tasks in the networking playbook has some caveats, let's sum up:

- Remove all networkd configuration
- Remove FAI default configuration
- Add 80-wired (since it’s just been removed…)
- Apply systemd-networkd with network_vars (including .link files) but **we do not reload networkd because the .link files have not yet been taken into account**
- First we apply the .link files (udev)
- Since all physical interfaces are ready (renamed or else…), we can create the cluster network. This configuration will create new interfaces (OVS bridge team0, etc.)
- Now that all interfaces exist, we can reload networkd
- Then we deploy the OVS topology and execute the ovs-vsctl extra commands
- Configuration of the /etc/hosts file
- DNS : Systemd-resolved configuration
- Time sync : Timemaster configuration
- systemd-networkd-wait-online.service configuration
- SRIOV libvirt network pool
- configuration of the guests br0 interfaces (including a networkd configuration reload)
- push iptables rules

You want to move the "configuration of the guests br0 interfaces" before we "apply systemd-networkd with network_vars" so that we can override whatever we like with the inventory, and I support the idea.

However "configuration of the guests br0 interfaces" will reload the networkd configuration, which is not acceptable before we apply the .link files. If a link file renames an interface, and the networkd configuration is written with the renamed interface in mind, reloading too soon may make some interfaces lose their IP configuration, while we are running a playbook remotely...

I suggest to remove this task.

insatomcat:
used problem solved with #415

PR #398, opened 2024-02-23
Make isolcpus and workqueues optional and revertable

insatomcat:
This commit makes the isolcpus and workqueues configurations optional.
If they were previously set up and and unconfigured later in the inventory, this commit also deals with the revert operations.

### DIFF
--- a/roles/debian/hypervisor/tasks/main.yml
+++ b/roles/debian/hypervisor/tasks/main.yml
@@ -103,12 +103,36 @@
     - cpufreq.default_governor=performance
     - intel_pstate=disable
     - nosoftlockup
-    - "rcu_nocbs={{ isolcpus }}"
     - rcu_nocb_poll
     - rcutree.kthread_prio=10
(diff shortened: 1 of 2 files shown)


roles/debian/hypervisor/tasks/main.yml
```diff
@@ -103,13 +103,37 @@
...
     - intel_pstate=disable
     - nosoftlockup
-    - "rcu_nocbs={{ isolcpus }}"
     - rcu_nocb_poll
     - rcutree.kthread_prio=10
-    - "isolcpus=nohz,domain,managed_irq,{{ isolcpus }}"
     - skew_tick=1
     - tsc=reliable
 
+- name: "grub conf hypervisor isolcpus"
```
dupremathieu, line 111: It works, but I think in the future we should create a file inside /etc/default/grub.d/ instead of modifying directly /etc/default/grub. It is easier to revert.

dupremathieu (approved):

PR #399, opened 2024-02-23
playbooks: deploy_vms_standalone: add gziped image support

lucbeaufils:
Some VM images can be archived using gzip to reduce the size during the deployment. When setting the ansible variable `disk_extract` to true, the image disk is copied and unarchived on the target.


playbooks/deploy_vms_standalone.yaml
```diff
@@ -34,7 +35,19 @@
...
       with_items: "{{ groups['VMs'] }}"
-      when: disk_copy | bool
+      when: disk_copy | bool and not(disk_extract | bool)
+    - name: Copy the gunziped disk on target
+      copy:
+        src: "{{ hostvars[item].vm_disk }}"
+        dest: "{{ disk_pool }}/{{ hostvars[item].inventory_hostname }}.img.gz"
+      with_items: "{{ groups['VMs'] }}"
+      when: disk_copy | bool and disk_extract | bool
+    - name: Extract the gunziped disk on target
```
dupremathieu, line 45: gzipped

playbooks/deploy_vms_standalone.yaml
```diff
@@ -34,7 +35,19 @@
         src: "{{ hostvars[item].vm_disk }}"
         dest: "{{ disk_pool }}/{{ hostvars[item].inventory_hostname }}.qcow2"
       with_items: "{{ groups['VMs'] }}"
-      when: disk_copy | bool
+      when: disk_copy | bool and not(disk_extract | bool)
+    - name: Copy the gunziped disk on target
+      copy:
+        src: "{{ hostvars[item].vm_disk }}"
+        dest: "{{ disk_pool }}/{{ hostvars[item].inventory_hostname }}.img.gz"
```
dupremathieu, line 42: why img.gz and no qcow2.gz?

playbooks/deploy_vms_standalone.yaml
```diff
@@ -34,7 +35,19 @@
...
+      with_items: "{{ groups['VMs'] }}"
+      when: disk_copy | bool and disk_extract | bool
+    - name: Extract the gunziped disk on target
+      command:
+        cmd: "gunzip {{ disk_pool }}/{{ hostvars[item].inventory_hostname }}.img.    gz"
```
dupremathieu, line 47: Remove spaces between img. and gz.

dupremathieu (changes requested):
The title is strange.
I will replace gunziped by gzipped because gunzip is for GNU unzip.
Add treatment is an unusual word for this meaning. We can replace it by support.

Note that the qcow2 file can be autocompressed if you want a small image.

Why don't you use the ansible.builtin.unarchive which supports a lot of compression formats and supports the decompression on the fly?

dupremathieu (dismissed):

dupremathieu (approved):

PR #402, opened 2024-02-26
Refactor the snmp logic

insatomcat:
The snmp logic is currently as follows everytime we need to expose something:
- we write a script specific to the need
- we modify the snmpd configuration to add an "extend" configuration item
- we add a sudo rule for the Debian-snmp account to be autorized to run the script

This commit propose a new approach, based on a "pass_persist" snmp agent configuration:
- a unique script (get_data.py) run as root will be charged to get all the data we need and write it in a single file (/tmp/snmpdata.txt)
- snmpd will only run a pass_persist script (exposeseapathsnmp.py) that will expose the content of the aforementionned file

With this approach:
- we don't need a specific new script for any new need
- we don't need sudo anymore, and so we don't need to change the sudo configuration when adding stuff to expose
- we do not need to touch the snmpd configuration when modifying what we want to expose
- we are in complete control of our snmp tree

Caveat: with pass_persist we cannot expose multiline strings


src/debian/snmp/snmp_getdata.py
```diff
@@ -0,0 +1,126 @@
+#!/usr/bin/python3
+
+import subprocess,json,os
+
+def run_command(command):
+    result = os.popen(command).read()
```
dupremathieu, line 6: How your script react if there is an error in the command?

dupremathieu (approved):
I like the idea.
Maybe we could improve the snmp_getdata.py script to log command errors.

src/debian/snmp/snmp_getdata.py
insatomcat: The run_command function will show the error:
```
# /usr/local/sbin/snmp_getdata.py
/bin/sh: 1: no: not found
```

and the oid will be empty in the /tmp/snmpdata.txt file but it won't impact the rest of the processing:
```
# head /tmp/snmpdata.txt
.2.1:smartctl /dev/sda
.2.1.0:PASSED
.2.2:smartctl /dev/sdb
.2.2.0:PASSED
.2.3:volontary error
.2.3.0:
.1.1:/usr/sbin/crm status
.1.1.1:Status of pacemakerd: 'Pacemaker is running' (last updated 2024-02-26 15:42:09 +01:00)
```

An empty value seems to be ignored by snmp_passpersist so it won't show in snmpwalk.

Anyway, I will had what is needed to log the program output to journalctl
```
*/5 * * * * root /usr/bin/systemd-cat -t "seapathsnmp-cron" /usr/local/sbin/snmp_getdata.py
```

PR #424, opened 2024-03-13
ci_vms_standalone_ptp.yaml: add vm ptp clock synchronisation in the CI

bchedotel:
Synchronize the system clock of each vms with phc2sys.

### DIFF
--- a/playbooks/ci_vms_standalone_ptp.yaml
+++ b/playbooks/ci_vms_standalone_ptp.yaml
@@ -1,0 +1,31 @@
+# Copyright (C) 2024 Savoir-faire Linux, Inc.
+# SPDX-Licence-Identifier: Apache-2.0
+
+# Ansible playbook that deploy and set-up ptp on vms for a standalone machine
+
+- import_playbook: deploy_vms_standalone.yaml
+
+- name: Check if phc2sys service is already started
+  hosts: VMs
+  tasks:
+    - name: Populate service facts
+      service_facts:
+
+    - name: Ensure phc2sys service is stopped
+      systemd:
+        name: phc2sys
+        state: stopped
+      when: "'phc2sys' in services"
+
+    - name: Copy systemd service file to server
+      copy:
+        src: ../src/phc2sys.service
+        dest: /etc/systemd/system
+        owner: root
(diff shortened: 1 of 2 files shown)


src/phc2sys.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Synchronize system clock or PTP hardware clock (PHC)
+Documentation=man:phc2sys
+After=ptp4l@%i.service
```
dupremathieu, line 4: You can not use the %i parameter with a regular service.
Create a template service: phc2sys@.service
```
[Unit]
Description=Synchronize system clock or PTP hardware clock (PHC) with %i
Documentation=man:phc2sys
Requires=ptp4l@%i.service
After=ptp4l@%i.service
Before=time-sync.target
```

dupremathieu (dismissed):

src/phc2sys.service
eroussy, line 4: ptp4l is simply not needed here, because we synchronise the VMs with ptp_kvm

eroussy (changes requested):

playbooks/ci_vms_standalone_ptp.yaml
```diff
@@ -0,0 +1,31 @@
...
+# SPDX-Licence-Identifier: Apache-2.0
+
+# Ansible playbook that deploy and set-up ptp on vms for a standalone machine
+
+- import_playbook: deploy_vms_standalone.yaml
+
+- name: Check if phc2sys service is already started
+  hosts: VMs
+  tasks:
+    - name: Gather facts about services
```
eroussy, line 11: I think you can just write 
```
    - name: Ensure phc2sys service is stopped
      systemd:
        name: phc2sys
        state: stopped
```
no need to gather facts are to type a when line

eroussy (changes requested):

src/phc2sys.service
bchedotel: I forget to remove the After, we do not use ptp4l because we use ptp_kvm inside the vm

eroussy (approved):

PR #429, opened 2024-03-19
roles/debian: disable Nginx if unused

TanguyRaufflet:
As it is already done with gunicorn.socket, disable Nginx if the VM manager API is not enable.


insatomcat:
nginx.service is supposed to always be disabled, even if the API is enabled.
https://github.com/seapath/ansible/blob/10e4b48ab2eb0201890f9233870279049bef1a44/roles/debian/physical_machine/tasks/main.yml#L76

AFAIK there is nowhere in seapath where the nginx service is enabled, so there was no need to disable it.

Maybe it's enabled by default from the debian installer, and we need to move this task after the block, so that it's applied even when the API is disabled ?

dupremathieu (approved):

insatomcat (changes requested):
nginx.service is supposed to always be disabled, even if the API is enabled.
https://github.com/seapath/ansible/blob/10e4b48ab2eb0201890f9233870279049bef1a44/roles/debian/physical_machine/tasks/main.yml#L76

AFAIK there is nowhere in seapath where the nginx service is enabled, so there was no need to disable it.

Maybe it's enabled by default from the debian installer, and we need to move this task after the block, so that it's applied even when the API is disabled ?

dupremathieu:
> nginx.service is supposed to always be disabled, even if the API is enabled.
> 
> https://github.com/seapath/ansible/blob/10e4b48ab2eb0201890f9233870279049bef1a44/roles/debian/physical_machine/tasks/main.yml#L76
> 
> AFAIK there is nowhere in seapath where the nginx service is enabled, so there was no need to disable it.
> 
> Maybe it's enabled by default from the debian installer, and we need to move this task after the block, so that it's applied even when the API is disabled ?

Debian enable and start systemd services by default.
With the default configuration, the service fails if the ipv6 is disabled.

I do not understand, where the Nginx service is enabled when we enable the API?

Do you confirm that if we move on top the Nginx disable part (and also stop the service) it will not break the vmmgr API?

insatomcat:
> I do not understand, where the Nginx service is enabled when we enable the API?

The nginx service is started by pacemaker, only on the host responding to the clusterIP 
https://github.com/seapath/ansible/blob/d02bbeefc6e898eff1952330cf4a7b4c17fe28fe/playbooks/cluster_setup_ha.yaml#L148


> @insatomcat do you confirm that if we move on top the Nginx disable part (and also stop the service) it will not break the vmmgr API?

We must move it outside of the block.
You can move it on top, but it seems cleaner to move it just next to the "disable gunicorn.service all the time" task.

insatomcat (changes requested):
We discussed about moving this task after the block, and you added it at the beginning.
Unless I have missed something, I still think moving (you can remove the original one) the task next to the gunicorn one is better.

dupremathieu:
> Unless I have missed something, I still think moving (you can remove the original one) the task next to the gunicorn one is better.

Are you sure we can do that? We want to disable and stop Nginx. The first time we run the playbook, it will work. If we re-run it a second time after Pacemaker has started Nginx the service will be stopped. Are you sure Pacemaker will restart the service?

insatomcat:
We only need to make sure it's disabled, not stopped.
I agree there is one corner case, the first time the playbook is run, nginx should be stopped and may be started since it's debian default.
I think the simplest way to deal with that is to disable it in build_debian_iso.

PR #435, opened 2024-03-20
cluster_setup_ceph: single disk ceph installation

TanguyRaufflet:
The ceph OSD disk is now zap only when a full disk is used for the ceph installation and the new
installation is now detected by the service
ceph-osd.target


TanguyRaufflet:
Draft PR

Issue #434, opened 2024-03-20
Add a branch verification  on startup


Paullgk:
**Is your feature request related to a problem? Please describe.**
Currently, there can be a confusion between using main branch or debian-main branch, as using main branch for configuring Seapath Debian and inversely. 

**Describe the solution you'd like**
A task in the beginning of configuration to make sure the correct branch is used with the correct Seapath flavor.
For example, with `uname -a` on Seapath Debian, we can detect we are using Seapath Debian:
`Linux machine1 6.5.0-0.deb12.1-rt-amd64 #1 SMP PREEMPT_RT Debian 6.5.3-1~bpo12+1 (2023-10-08) x86_64 GNU/Linux`

**Describe alternatives you've considered**
A  subsection has been added in README file: https://github.com/seapath/ansible/pull/432 and https://github.com/seapath/ansible/pull/433

Paullgk:
For main branch, fixed with #446

PR #440, opened 2024-03-20
cluster_setup_debian: add a branch verification

Paullgk:
Add a task to verify we are configuring the target with a compatible Ansible git branch.  #434

### DIFF
--- a/playbooks/cluster_setup_debian.yaml
+++ b/playbooks/cluster_setup_debian.yaml
@@ -6,6 +6,22 @@
 # with an iso created by build_debian_iso
 # This playbooks can also configure VMs created with build_debian_iso
 
+- name: Check if we are using a correct seapath flavor
+  hosts:
+    - cluster_machines
+    - standalone_machine
+    - VMs
+  tasks:
+    - name: Retrieve Seapath system version
+      command: "uname -a"
+      register: "uname_output"
+
+    - name: Check if current Ansible branch is compatible
+      fail:
+        msg: "The current branch (debian-main) is incompatible with the Yocto version of SEAPATH. Please use the main branch to configure this machine"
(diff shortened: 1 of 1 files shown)


playbooks/cluster_setup_debian.yaml
```diff
@@ -6,6 +6,21 @@
 # with an iso created by build_debian_iso
 # This playbooks can also configure VMs created with build_debian_iso
 
+- name: Check if we are using a correct seapath flavor
+  hosts:
+    - cluster_machines
+    - standalone_machine
```
eroussy, line 12: Add VMs in hosts

playbooks/cluster_setup_debian.yaml
```diff
@@ -6,6 +6,21 @@
...
+  tasks:
+    - name: Retrieve Seapath system version
+      command: "uname -a"
+      register: "uname_output"
+
+    - name: Check if current Ansible branch is compatible
+      fail:
+        msg: "You are trying to configure with a Seapath flavor incompatible with the current Ansible git branch"
```
eroussy, line 20: I would be more straight forward in the message :
"The current branch (debian-main) is incompatible with the Yocto version of SEAPATH. Please use the main branch to configure this hypervisor"

eroussy (changes requested):

playbooks/cluster_setup_debian.yaml
Paullgk: Done

playbooks/cluster_setup_debian.yaml
Paullgk: Done

eroussy (approved):

PR #444, opened 2024-03-21
Replace direct Real-Time tuning with Tuned.

yairpod:
We have a lot of code for the purpose of tuning the host machines for real-time Latency.
The TuneD is a system tuning service can simplify the code and improve results.
In my development test for this PR I measured the same cluster before this change and after,
By making a clean install from scratch and running a 12 hour cyclic test on each node on all isolated cpus.
results are:
Before:

```
* Node1 
	Min Latencies: 00003 00004 00003 00003 00003 00003 00004 00004 00003
	Avg Latencies: 00005 00007 00005 00005 00005 00005 00007 00007 00006
	Max Latencies: 00048 00015 00011 00027 00012 00027 00102 00016 00018
* Node2 
	Min Latencies: 00003 00005 00003 00003 00003 00003 00003 00005 00003
	Avg Latencies: 00005 00007 00004 00005 00005 00005 00005 00007 00005
	Max Latencies: 00017 00016 00012 00021 00017 00011 00030 00063 00020
* Node3 
	Min Latencies: 00003 00004 00004 00003 00003 00003 00003 00004 00003
	Avg Latencies: 00005 00007 00007 00005 00005 00005 00005 00007 00005
	Max Latencies: 00036 00015 00027 00012 00024 00011 00021 00015 00025
Max Latency 102
```
After:
```
* Node1 
	Min Latencies: 00003 00004 00004 00003 00003 00004 00004 00004 00004
	Avg Latencies: 00005 00005 00004 00007 00004 00004 00005 00005 00005
	Max Latencies: 00016 00014 00011 00027 00010 00030 00014 00018 00020
* Node2
	Min Latencies: 00003 00004 00003 00003 00003 00003 00003 00005 00005
	Avg Latencies: 00005 00007 00004 00004 00005 00005 00005 00007 00006
	Max Latencies: 00015 00015 00012 00026 00015 00012 00031 00015 00028
* Node3
	Min Latencies: 00003 00005 00005 00004 00003 00003 00003 00003 00003
	Avg Latencies: 00006 00007 00007 00007 00005 00005 00005 00007 00005
	Max Latencies: 00018 00018 00015 00014 00028 00011 00028 00016 00019
Max Latency 31
```

This would require adding Tuned to the Debian image, Please see PR:
https://github.com/seapath/build_debian_iso/pull/91


insatomcat (dismissed):

roles/debian/hypervisor/tasks/main.yml
```diff
@@ -179,21 +109,6 @@
...
-    line: "CPUAffinity={{ cpusystem }}"
-    state: present
-  when: cpusystem is defined
-- name: "systemd conf CPUAffinity revert"
-  lineinfile:
-    dest: /etc/systemd/system.conf
-    regexp: '^#?CPUAffinity=.*$'
-    line: "#CPUAffinity="
-    state: present
-  when: cpusystem is not defined
```
eroussy, line 195: Just a question here : 
How does Tuned interacts with the cgroups defined of SEAPATH Debian.
For example, here, you remove the CPUAffinity of the system slice, because it will be defined by Tuned.
But what about the other slices ? Like `machine.slice` for example ? Is it also handled by Tuned or do we still need to handle that externally ?

insatomcat: systemd slices are not managed with tuned at all

eroussy: Oh, I confuse `system.conf` and the `system.slice`, my bad.

But anyway, would it be possible to manage slices with tuned ?
For example here, it seems to me that we defined the same thing twice : 
- `isolcpus` is used in `src/tuned/seapath-rt-host-variables.conf.j2`
- `cpusystem` is used in `system.slice` configuration.
To me, these two variables will always be the same, so why not define them in the tuned configuration

insatomcat: They are not the same. These are different approaches: isolcpus vs cgroups/cpusets.

Tuned will not deal with slices/cgroups, so a standard use case will be to use tuned and not use slice at all. That way you isolate some cores, and use them in guests.
In that usecase you only have "isolated cores" and "non isolated cores", all slices have the same cpuaffinity (all cpus, but since you have isolcpus=domain... only "non isolated cores"), and you won't use cpusystem, cpuuser, cpumachines, etc. 

You can then choose you want to do housekeeping on specific cores, or run guests with a specific cpu affinity, in which case you will use cpusystem, cpumachinesrt, etc.
Like we discussed before, moving to slices/cpusets is linked to the isolcpus feature being deprecated, and seapath anticipating the future, but it's all optionnal for now.

This PR is really only about moving the isolcpus logic (with sysctl, sysfs, workqueues, etc.) to tuned. It does not involve the slices configuration which will still be managed by ansible (if used at all).

eroussy: Ok, thanks for the explanations

src/tuned/seapath-rt-host-variables.conf.j2
```diff
@@ -0,0 +1 @@
+isolated_cores={{ isolcpus | default('') }}
```
eroussy, line 1: Why make that in a separate template file ?
It is simpler to include this line in `src/tuned/tuned.conf` and templatize the entire file at once.

eroussy (changes requested):

src/tuned/seapath-rt-host-variables.conf.j2
insatomcat: I guess it's just the usual way of tuned doing things. Usually the profiles are static (in /usr...) and customizables with variables in /etc
That's how @yairpod first submitted the PR, by only modifying a variable file.
Since we created our own Seapath profile, we did the same things, a "static" profile in tuned.conf and variables in a separate file.

insatomcat: I merged both files

dupremathieu (approved):

eroussy (dismissed):

dupremathieu (approved):

insatomcat (approved):

PR #445, opened 2024-03-21
Correct standalone deployement

eroussy:


Paullgk (changes requested):
Commit message is weird: 

```
 I think it is better to create a new playbook than to put a when clause
in `cluster_setup_debian` and use it for standalone.
```

Paullgk (commented):
Does the branch verification is done with the standalone playbook?

Paullgk (approved):

PR #451, opened 2024-03-22
cluster_setup_ceph: ceph partition creation for cluster

TanguyRaufflet:
Modifications concerning the creation of the ceph partition in the case of a single disk available


insatomcat:
can you please explain the problem you are trying to solve?

TanguyRaufflet:
I'm trying to install a cluster of 3 nuc. Initially, the configuration of Ceph was designed to work with a second disk dedicated to it. Since, I only have one disk for both Ceph and Linux, I have to create a partition in the same disk as the one linux is running.

insatomcat:
There're already everything you need in the current playbook.
You can solve your issue 2 ways:
- create manually the ceph VG and LV 
- have the playbook create them (and it can even create the partition)

in both case, something like that will do:
```
lvm_volumes: 
  - data: lv_ceph # Name of the logical volume to use for the CEPH OSD volume
    data_vg: vg_ceph # Name of the volume group to use for the CEPH OSD volume
    data_size: 5G # Size of the logical volume, default in megabytes, possible values: [0-9]+[bBsSkKmMgGtTpPeE]
    device: /dev/sda #the disk you want to create the partition on
    device_number: 5 # Number of the partition in the lvm_volumes[0].device
    device_size: 7GiB # Size of the partition in lvm_volumes[0].device, default in megabytes, possible units: B, KB, KiB, MB, MiB, GB, GiB, TB, TiB
```

TanguyRaufflet:
yes you right, I'm using lvm_volumes and it's supposed to work. But when I run the playbook cluster_setup_ceph from the debian-main branch, I have the following error :
Here the partitions are already created and ceph-osd is already running. So, he tries
```
fatal: [nuc1]: FAILED! => changed=true 
  cmd:
  - ceph-volume
  - lvm
  - zap
  - /dev/vg_ceph/lv_ceph
  delta: '0:00:00.359108'
  end: '2024-03-25 14:59:40.010792'
  msg: non-zero return code
  rc: 1
  start: '2024-03-25 14:59:39.651684'
  stderr: |-
    --> Zapping: /dev/vg_ceph/lv_ceph
    --> Unmounting /var/lib/ceph/osd/ceph-1
    Running command: /usr/bin/umount -v /var/lib/ceph/osd/ceph-1
     stderr: umount: /var/lib/ceph/osd/ceph-1: target is busy.
    -->  RuntimeError: command returned non-zero exit status: 32
  stderr_lines: <omitted>
  stdout: ''
  stdout_lines: <omitted>
```
I've even try to re-flash one of the nuc to have a clean install and it gives me an error related to the fact that the disk is already used.

TanguyRaufflet:
Yes you are right, I'm using lvm_volumes and it's supposed to work. But when I run the playbook taken from the branch debian-main with a nuc I just re-flash, I'm getting the following error :
```
fatal: [nuc3]: FAILED! => changed=true 
  cmd:
  - ceph-volume
  - lvm
  - zap
  - /dev/nvme0n1
  - --destroy
  delta: '0:00:00.454673'
  end: '2024-03-25 15:41:03.147799'
  msg: non-zero return code
  rc: 1
  start: '2024-03-25 15:41:02.693126'
  stderr: '-->  RuntimeError: Device /dev/nvme0n1 has partitions.'
  stderr_lines: <omitted>
  stdout: ''
  stdout_lines: <omitted>
```

insatomcat:
ok, by reading the code (it's a long time...) I understand the case "create the partition then the PV, then the VG then the LV..." only works on a seperate disk. That's not what I had in mind.
You can still solve your problem by creating the VG/LV manually on same disk.
But if you prefer to make it work by modifying the playbook, it's fine with me :)
Reminder: the playbook is supposed to deal with 4 different cases:
- using LVM with pre-existing LV/VG
- using LVM and create what's missing (partition, PV, VG, LV...)
- not using LVM -> use a whole disk
Thanks.

playbooks/cluster_setup_ceph.yaml
```diff
@@ -60,11 +60,12 @@
...
                   - name: Create partition on OSD disks with parted
                     community.general.parted:
-                      device: "{{ lvm_volumes[0].device }}"
+                      device: "{{ ceph_osd_realdisk.stdout }}"
                       number: "{{ lvm_volumes[0].device_number }}"
+                      label: gpt # Must be the same type as the partition table of the disk used
                       state: present
                       unit: MB
-                      part_end: "{{ lvm_volumes[0].device_size }}"
+                      part_start: "-{{ lvm_volumes[0].device_size }}" # Negative numbers specifie the distance from the end of the disk
```
Paullgk, line 68: In this case, you are allocating your new partition with the total size of disk available?
Did you check if the filesystem needs to be refreshed?

Paullgk (changes requested):

playbooks/cluster_setup_ceph.yaml
TanguyRaufflet: No the size of the new partition is defined by device_size.
I don't see why a refresh of the filesystem should be done.

Paullgk (approved):

PR #453, opened 2024-03-26
cluster_setup_network: making cluster_ip_addr's variable optionnal

TanguyRaufflet:
It is no longer mandatory to define the variable cluster_ip_addr in the inventory in the case we don't need it. That is needed according to 3f7d138


insatomcat:
unless I'm mistaken the commit https://github.com/seapath/ansible/commit/3f7d13826929889f5706b27d1a8a904aa0423b12 was correcting an issue with a standalone deployment.
cluster_ip_addr is mandatory for a cluster deployment, and the task you modified is the task for cluster_machines.

There is another "Build hosts file" task just below that will deal with standalone_machine and that does not require cluster_ip_addr

TanguyRaufflet:
In my case, the cluster is made through a switch. So, if I'm not mistaken, it's normal to don't have the variable cluster_ip_addr defined in this specific case.

insatomcat:
I'm not sure I see why. Each node must have its address on the cluster_network, that's the role of the cluster_ip_addr.

TanguyRaufflet:
In my case, there is no cluster network, the nodes are linked through their ansible_host's ip.

insatomcat:
I don't think this is a usecase we want to support or complicate the codebase for. A seapath cluster needs a resillient cluster network (a single switch is not resilient) and so we only support a loop topology with HSR or RSTP for now.
The network playbook will try to create the cluster network anyway with a team0 bridge and 2 physical interface (how do you deal with that if you do not have physical nic to dedicate to this?)

What you can try is to overide the network configuration (with custom_network and custom_netdev variables), so that you create your own cluster network as secondary addresses of your "remote access" (ansible_host) network. This would not require to change the codebase since it would entirely be dealt with in your inventory.

dupremathieu:
Our use case is to be able to have a basic SEAPATH cluster based on retail PCs such as Intel NUC for quick demos or to make it easier for a newcomer to get started with SEAPATH. I agree in production a cluster network is required.

TanguyRaufflet:
What do you think @insatomcat ?

insatomcat:
I don't really like the idea of complexifing the codebase just to make a new user's experience simpler.
New users is ephemeral, maintaining the code is not...
However I understand the need, Seapath being in an early adoption phase, so I won't oppose it.
But please since this is unintuitive, please make sure this is properly documented (in the code, and in the commit) so that a new developer will understand why we made that change.

playbooks/cluster_setup_network.yaml
```diff
@@ -173,14 +173,22 @@
...
-    - name: Build hosts file
+    - name: Build hosts file (Cluster Network)
       lineinfile:
         dest: /etc/hosts
         regexp: '.*{{ item }}(| rbd)$'
         line: "{{ hostvars[item].cluster_ip_addr }} {{ item }}{% if item == inventory_hostname %} rbd{% endif %}"
         state: present
       when: hostvars[item].cluster_ip_addr is defined
       loop: "{{ groups['cluster_machines'] }}"
+    - name: Build hosts file (Public Network)
```
dupremathieu, line 184: You can add a comment to mention that if cluster_ip_addr is not defined, we consider that there is no cluster Network, and we use the public network for Ceph instead.

dupremathieu (changes requested):

playbooks/cluster_setup_network.yaml
TanguyRaufflet: Done

dupremathieu (approved):

PR #458, opened 2024-03-28
networkd-wait: allows the user to customize the interface to wait for

insatomcat:
By default, systemd-networkd-wait-online will wait for br0 (or the vlan on br0 if using br0vlan variable) and the cluster_network interface (team0 or hsr0).
In some setups it is useful to customize the network_interface the most relevant to wait for at boot time.
This new variable will override "br0" as the name of the interface we will wait for.

### DIFF
--- a/inventories/seapath_cluster_definition_example.yml
+++ b/inventories/seapath_cluster_definition_example.yml
@@ -313,6 +313,8 @@ all:
           INTERFACE3: [1-4094]  # no filtering, excluding untagged trafic
           INTERFACE4: []        # no filtering, including untagged trafic
 
+        #interface_to_wait_for: eth3 #by default, systemd-networkd-wait-online will only wait for the cluster_network interface (team0 or hsr0). In some setup it is useful to customize the network_interface the most relevant to wait for at boot time. This variable will add an interface networkd will wait for at boot time
+
         iptables_rules_path: "../inventories/iptables_rules_example.txt" #when defined, it uploads this file to /etc/iptables/rules.v4 and loads those rules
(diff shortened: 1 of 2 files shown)


eroussy (commented):
The code seems correct to me, but I fail to see the use of it.
Could you give an example of case where this is useful ?

Regarding the SEAPATH cluster, the most important interface to wait for is team0 (or hrs) 
Waiting for br0 to be up is useful for the VM, I can understand that. 
But in which case do we need one more interface ?

insatomcat:
> The code seems correct to me, but I fail to see the use of it. Could you give an example of case where this is useful ?
> 
> Regarding the SEAPATH cluster, the most important interface to wait for is team0 (or hrs) Waiting for br0 to be up is useful for the VM, I can understand that. But in which case do we need one more interface ?

you are right but this is not directly the point of this PR.
Yes, we can question the need to wait for something else than br0, this PR is about changing the br0 into something more specific.
The usecase is real: on one setup we don't use the name "br0" and we customize the names of all network interfaces (using custom_network/netdev/link variables), so when the server boots, it waits for an interface that does not exist.

However I can propose something that will solve both issues:
```
ExecStart=/lib/systemd/systemd-networkd-wait-online {{ "-i " + interface_to_wait_for if interface_to_wait_for is defined }}{% if cluster_protocol is defined and cluster_protocol == 'HSR' %} -i hsr0 {% elif cluster_ip_addr is defined %} -i team0 {% endif %}
```

That way we only wait for the cluster network interface, unless the user specifically wants to add an another interface to wait for.
I updated the commit, tell me what you think.

eroussy (approved):

PR #460, opened 2024-03-29
Harmonize livemigration debian

eroussy:
This PR backport the playbook cluster_setup_add_livemigration_user.yaml from yocto.
It aims to harmonize the creation of this user between yocto and debian


playbooks/deploy_vms_cluster.yaml
```diff
@@ -21,8 +21,9 @@
         command: create
         system_image: /tmp/os_{{ item }}.qcow2
         force: true
-        live_migration: true
+        live_migration: "{{ true if livemigration_user is defined }}"
```
insatomcat, line 24: shouldn't this condition be on the line just after this one?

eroussy: No.
`live_migration` define if we want to use livemigration or not.
`migration_user` define the user to do it.
In the case of our ansible inventory, live migration is active only if we define a user for it. So only if the variable `livemigration_user` is defined.

But I forget to explain that in the ansible inventory. I do it

insatomcat: ok but maybe this condition should also be on the migration_user line below.
If livemigration_user is not defined, won't the playbook fail otherwise?

inventories/seapath_cluster_definition_example.yml
```diff
@@ -147,21 +147,6 @@ all:
...
                         cpuovs: "23" # CPUs reserves for OVS
 
-        # VM group: use to deploy and manage guests on the cluster
-        VMs:
-            hosts:
-                test:
-                  live_migration: true
-                  preferred_host: "node3"
-                  migrate_to_timeout: 180
-                  enable: false
```
Paullgk, line 158: Did you add this removed part in `seapath_vm_definition_example.yaml`? I don't see it anywhere in the file or in your PR

Paullgk (changes requested):

inventories/seapath_cluster_definition_example.yml
eroussy: No, because `seapath_vm_definition_example.yaml` is supposed to work with the template `guest.xml`.
The VM that I removed here is supposed to work with the seapath_rt and seapath_nort VM xml.
We should harmonize these xml to be more clear, but I will not do it in this PR.

Paullgk (approved):

PR #468, opened 2024-04-05
Inventories move cgroup to readme

eroussy:
Move the description of the cgroups in the README instead of directly in the inventories

### DIFF
--- a/inventories/README.md
+++ b/inventories/README.md
@@ -16,25 +16,13 @@ It is possible with Ansible to provides [multiple inventories files](https://doc
 * the physical machines
 * the cluster internal network (Open vSwitch topology)
 * the virtual machines
-## SEAPATH Ansible inventories examples description
 
-All inventories examples are in YAML format. The example inventories are:
-* `seapath_demo_cluster_definition_example.yml`: SEAPATH cluster demo
-* `seapath_standalone_definition_example.yml`: SEAPATH standalone hypervisor setup
-* The complex SEAPATH cluster inventories are:
-	* `seapath_cluster_definition_example.yml`: physical machines
-	* `seapath_ovstopology_definition_example.yml`: cluster internal network (Open vSwitch topology)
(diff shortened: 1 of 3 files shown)


inventories/seapath_cluster_definition_example.yml
```diff
@@ -136,16 +136,9 @@ all:
                         #OPTIONAL
                         logstash_server_ip: 10.10.10.10
 
-                        # Realtime/CPUs cgroups isolation configuration
-                        # All those parameters are optional.
-                        isolcpus: "2-5,14-17" # CPUs to isolate
-
-                        cpusystem: "0,12" # CPUs reserves for system
```
dupremathieu, line 143: Why not keep this variable in the example?
To me, the idea of these inventories is to give a usage example of all the possible variables. Do you want to change this?

eroussy: I agree that these inventories must provide usage examples.
However, they are also a starting point for people that want to implement SEAPATH : They copy this inventory and modify it line by line.
I think it is a good idea to keep this inventory simple for those people.

Particularly, these cpu variables are advanced variables and should not be proposed to beginners.
I think the best would be to have different kind of inventories (basic, networking, cgroups, etc ...) for examples, but I will not do that now.

dupremathieu: We have the `seapath_demo_cluster_definition_example.yml` inventory for beginners.

inventories/README.md
```diff
@@ -16,25 +16,13 @@ It is possible with Ansible to provides [multiple inventories files](https://doc
 * the physical machines
 * the cluster internal network (Open vSwitch topology)
 * the virtual machines
-## SEAPATH Ansible inventories examples description
```
dupremathieu, line 19: I preferred keeping SEAPATH Ansible inventories examples description section before Advanced variables.

PR #477, opened 2024-04-17
ptpstatus: reimplementation our own script

### LINKED ISSUE #476: License Violation in ptpstatus.sh  (debian-main branch)
The License described in this file is problematic for usage with ansible.
https://github.com/seapath/ansible/blob/debian-main/src/debian/ptpstatus/ptpstatus.sh

dupremathieu:
To avoid copyright issues, we reimplemented the ptpstatus script. It provides the same output files but, it is less verbose. The refresh interval is now a parameter of the script.

Fix #476


eroussy (changes requested):
commit message title : "reimplement our own script"

otherwise, LGTM

eroussy (approved):

PR #483, opened 2024-04-24
Autodetect vm main disk backport for Yocto

eroussy:
The main qcow2 file is now automatically added to the local_disk list.
 It allows the VM to boot without having to declare the disk in the VM
 inventory.


kevlhop (approved):

eroussy:
Closing this PR as not supported for now.
Because the Yocto version is on ansible 2.9, this modification of the playbook fails.

This will be done when merging the branches debian and yocto

PR #506, opened 2024-06-03
introducing conntrackd for conntrack table synchronization

insatomcat:
Adds optional conntrackd synchronization feature.
This is useful when using the hosts as routers (when using iptables for flow security between bridges), if you want the conntrack tables to be synchronized among hosts. This is necessary because when the default gateways for the routed networks migrate, the connection tracking status is lost.
If conntrackd_ip_list is defined it will enable the appropriate portion of the playbook, and the content will be used as the list of IP addresses to enable conntrack synchronization on.

### DIFF
--- a/roles/debian/hypervisor/tasks/main.yml
+++ b/roles/debian/hypervisor/tasks/main.yml
@@ -272,3 +272,10 @@
     name: ptp_vsock.service
     enabled: yes
     state: started
+
+- block:
+  - name: make sure conntrackd package is installed
+    ansible.builtin.apt:
+      name: conntrackd
+      state: present
+  when: conntrackd_ip_list is defined
--- a/src/conntrackd/conntrackd.conf.j2
+++ b/src/conntrackd/conntrackd.conf.j2
@@ -1,0 +1,50 @@
+Sync {
+        Mode NOTRACK {
+                DisableInternalCache on
+                DisableExternalCache on
+                StartupResync on
+        }
+        Multicast {
+                IPv4_address 225.0.0.50
+                Group 3780
+		IPv4_interface {{ cluster_ip_addr }}
+                Interface team0
+                SndSocketBuffer 1249280
+                RcvSocketBuffer 1249280
+                Checksum on
+        }
+        Options {
+                TCPWindowTracking Off
+                ExpectationSync On
+        }
+}
+General {
+        Systemd on
+        HashSize 32768
+        HashLimit 131072
+        Syslog on
+        LockFile /var/lock/conntrackd.lock
+        UNIX {
+                Path /var/run/conntrackd.sock
+        }
+        NetlinkBufferSize 2097152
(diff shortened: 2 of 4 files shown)


roles/debian/hypervisor/tasks/main.yml
```diff
@@ -272,3 +272,29 @@
     name: ptp_vsock.service
     enabled: yes
     state: started
+
+- block:
```
eroussy, line 276: This should not be in the debian specific part. Could you move it in `playbooks/cluster_setup_network.yaml` ?

roles/debian/hypervisor/tasks/main.yml
```diff
@@ -272,3 +272,29 @@
...
+  - name: make sure conntrackd package is installed
+    ansible.builtin.apt:
```
eroussy, line 278: I would prefer checking that the systemd service `contrackd` exists, rather than the apt package.

src/conntrackd/conntrackd.conf.j2
```diff
@@ -0,0 +1,52 @@
...
+        NetlinkBufferSizeMaxGrowth 8388608
+        NetlinkOverrunResync On
+        NetlinkEventsReliable Off
+        EventIterationLimit 100
+        Filter From Kernelspace {
+                Protocol Accept {
+                        TCP
+                }
+                Address Accept {
+                        {% if conntrackd_ip_list is defined -%}
```
eroussy, line 40: Is there really a need to check for that ?
This file should not be uploaded if the variable `conntrackd_ip_list` doesn't exists.

eroussy (changes requested):

roles/debian/hypervisor/tasks/main.yml
insatomcat: for the CI to pass this PR, the package needs to be installed. But if the package is installed, then the cukinia-tests will fail for other PR (the package is not supposed to be there).
The workaround is to have the package installed by apt for the time being.
Once the PR is merged, we can add the package to build_debian_iso, change the cukinia-test, and then we can remove this part altogether.

roles/debian/hypervisor/tasks/main.yml
insatomcat: same answer as before, this is a workaround to install the package on the fly, while the change is not merged

src/conntrackd/conntrackd.conf.j2
insatomcat: that's fair, will change

roles/debian/hypervisor/tasks/main.yml
eroussy: Okay, but please don't forget to remove this apt part later and to move that in the network playbook.

eroussy (dismissed):

eroussy (approved):

PR #519, opened 2024-06-24
scripts/ci_system_info.py: add system info fetch

bchedotel:
Addition of a script to retrieve information
from the system on which the CI will run.

### DIFF
--- a/playbooks/ci_all_machines_tests.yaml
+++ b/playbooks/ci_all_machines_tests.yaml
@@ -4,7 +4,10 @@
 # Ansible playbook that runs all Cukinia's tests.
 
 ---
-- hosts: "{{ machines_tested | default('cluster_machines') }}"
+- hosts: 
+    - cluster_machines
+    - standalone_machine
+    - VMs
   name: Cukinia tests
   tasks:
     - name: Create temporary Cukinia directory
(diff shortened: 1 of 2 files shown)


playbooks/ci_system_info.py
```diff
@@ -0,0 +1,43 @@
...
+                _cpu_arch_ = platform.machine(),
+                _cpu_log_ = psutil.cpu_count(),
+                _ram_ = str(round(psutil.virtual_memory().total / (1024.0 **3)))+" GB" 
+            )
+    )
+
+    for i in range(len(psutil.net_if_stats())):
+        adoc_file.write(
+            network_template.format(
+                _net_if_name_ = list(psutil.net_if_stats())[i],
```
eroussy, line 38: That display all the interfaces including loopback, docker or internal bridges.
It would be better to display only physical interfaces.
It would be even better if you can display what network card it is and what capabilities it have.

playbooks/ci_system_info.py
```diff
@@ -0,0 +1,43 @@
+import platform
```
eroussy, line 1: You need to call this file inside a playbook, do that here in  the same pull request.

eroussy, line 1: This is a script, not a playbooks. Put this file in the script directory.

eroussy (changes requested):

playbooks/ci_system_info.py
bchedotel: Couldn´t we just modify the CI script to run this script ?

playbooks/ci_all_machines_tests.yaml
```diff
@@ -23,4 +23,21 @@
         dest: >-
           {{ cukinia_test_prefix | default('..') }}/cukinia_{{ inventory_hostname
           }}.xml
+        flat: yes       
+
+- hosts: "{{ machines_tested | default('cluster_machines') }}"
```
eroussy, line 28: Is it not better to apply that on "cluster_machines" only ?
Otherwise, you can target machine tested without VMs with
`hosts: "{{ (machines_tested | default(groups['cluster_machines'])) | difference(groups['VMs']) }}"`

eroussy (changes requested):

eroussy (approved):

PR #521, opened 2024-06-25
ansible.cfg: set gathering to explicit

dupremathieu:
False is not a valid value for the gathering option. The correct value is explicit. Ansible version greater than 2.10 will raise an error if the value is set to False.


ebail (approved):
Thanks !

PR #528, opened 2024-06-28
openlab: add cyclic tests

bchedotel:
Add cyclic testing during the
CI tests.

### DIFF
--- a/scripts/run_cyclictest.py
+++ b/scripts/run_cyclictest.py
@@ -1,0 +1,15 @@
+#!/usr/bin/env python3
+
+import subprocess
+import sys
+import argparse
+
+parser = argparse.ArgumentParser(description="Run cyclictest command.")
+parser.add_argument('output_file', type=str, help='File to output the result.')
+args = parser.parse_args()
+
+cyclic_test_cmd = "cyclictest -l100000 -m -Sp90 -i200 -h400 -q"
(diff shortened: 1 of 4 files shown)


scripts/run_cyclictest.py
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env python3
+
+import subprocess
+import sys
+
+if len(sys.argv) != 2:
+    print("Usage: ./run_cyclictest.py <output_file>")
+    sys.exit(1)
```
Paullgk, line 8: Use argparse

Paullgk (changes requested):

scripts/run_cyclictest.py
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env python3
+
+import subprocess
+import sys
+import argparse
+
+parser = argparse.ArgumentParser(description="Run cyclictest command.")
+parser.add_argument('output_file', type=str, help='File to output the result.')
```
Paullgk, line 8: If these arg is not provided, this will probably fail. Add required=true

Paullgk (changes requested):

scripts/run_cyclictest.py
bchedotel: I created a positional argument (no -- option in front of the name). Positional arguments are always required so I cannot use required=True for such options

Paullgk (approved):

PR #539, opened 2024-07-02
Add hosts file upload to cluster and standalone architecture and iptables rules feature to standalone architecture

Gabin-rte:
Add iptables feature to standalone

The iptables feature that allows the seapath user to use a template or a static file to load iptables rules was not available for standalone architecture. This fixes this issue.

----
Add option to upload user hosts file

A hosts file can be provided to add local name resolution. Indeed there will most probably be no DNS in a substation.
This file is specified using the hosts_path variable and can be either a plain text file or a Jinja template.


insatomcat (dismissed):

insatomcat (commented):
LGTM

dupremathieu (approved):

PR #545, opened 2024-07-04
playbooks/tasks/deploy_vm.yaml: change default value

dupremathieu:
* vm_file can also be defined in the VM hostvars
* vm_file_dest have a default value: /tmp/os_[vm name].qcow2
* disk_copy is true by default
* live_migration is false by default instead of empty string

### DIFF
--- a/playbooks/tasks/deploy_vm.yaml
+++ b/playbooks/tasks/deploy_vm.yaml
@@ -1,4 +1,5 @@
 # Copyright (C) 2021, RTE (http://www.rte-france.com)
+# Copyright (C) 2024 Savoir-faire Linux, Inc
 # SPDX-License-Identifier: Apache-2.0
 # Task to deploy a VM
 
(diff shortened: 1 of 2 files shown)


playbooks/tasks/deploy_vm.yaml
```diff
@@ -24,18 +24,18 @@
 - block:
       - name: "Copy {{ item }} system disk on target"
         copy:
-          src: "{{ vm_file }}"
-          dest: "{{ vm_file_dest }}"
+          src: "{{ hostvars[item].vm_file | default(vm_file) }}"
```
eroussy, line 27: This is already done in `deploy_vms_cluster.yaml` playbook

eroussy (changes requested):

playbooks/tasks/deploy_vm.yaml
eroussy: okay, but you have to put the behaviour of `vms_disks_directory` in this task now : 
`src: "{{ hostvars[item].vm_disk | default( (vms_disks_directory|default('')) ~ '/' ~ item ~ '.qcow2') }}"`
This is useful to be consistent with the debian branch

playbooks/tasks/deploy_vm.yaml
```diff
@@ -24,18 +24,18 @@
...
+          dest: "{{ vm_file_dest | default ('/tmp/os_' + item + '.qcow2') }}"
```
eroussy, line 28: same here : 
`dest: "{{ qcow2tmpuploadfolder | default('/tmp') + '/os.qcow2' }}"`
and in all other mentions of the "vm_file_dest" variable

eroussy (changes requested):

PR #546, opened 2024-07-05
Sync network with debian

eroussy:
Sync the network configuration with debian.
The default configuration is yet the same as debian (br0 on main interface). This prepares the jump to netplan.

Also add various become: true in the playbooks to be able to use ansible on an unprivileged user.


dupremathieu (changes requested):
Update copyrights

dupremathieu (approved):

PR #548, opened 2024-07-05
playbooks/cluster_setup_ceph: disable unsurported / unwanted modules

dupremathieu:
The cephadm, prometheus and rook modules are not supported by the SEAPATH Yocto.

The dashboard and restful modules should be disabled if the dashboard is not enabled.


dupremathieu:
In the end, we do not install the unwanted plugins, so no need to disable it.

PR #558, opened 2024-07-12
cleanup: remove obsolete files

insatomcat:
- examples folder: all this is dealt with by vm_manager and the deploy_vm playbooks.
- vm_templates: they are now in /templates/vm
- module_documentation: the ansible module documentation is published on ansible-galaxy (seapath.cluster_vm)
- roles: seapath only supports one-disk vm + all the rest is dealt with by vm_manager and the deploy_vm playbooks
- tests/tests_results: all this is dealt with by cukinia / cukinia-tests and the CI
- vm_images: this was for an example, the user can create the folder if he needs


vm_images/README
```diff
@@ -1,2 +0,0 @@
-VM disk images belong in this folder.
```
eroussy, line 1: This will remove the directory `vm_images` from the git repo.
Edit the main README and the inventories examples to reflect this change or add a .gitkeep

module_documentation_sources/plugin_formatter.py
```diff
@@ -1,782 +0,0 @@
-# Copyright: (c) 2012, Jan-Piet Mens <jpmens () gmail.com>
```
eroussy, line 1: removing this `module_documentation_sources` is a good idea, but then we don't have any documentation for the ansible plugin anymore.
The best would be to upload the plugin to ansible-galaxy. If the documentation is well done (it should already be the case), it is automatically displays on the site of ansible galaxy.
what do you think ?
(also we should update the README)

eroussy (changes requested):

vm_images/README
insatomcat: I edited the main README to indicate the user that he needs to create the folder first

module_documentation_sources/plugin_formatter.py
insatomcat: like you said, the documentation exists, it is embedded in the module code + it's easy to generate it with ansible-doc (text mode) + the user can generate it himself in html format with sphinx. Right now we do not provide the documentation, we provide a broken sphinx script to help the user create it... So I don't think it will be missed.
However I find your idea of uploading the module to ansible-galaxy very good.
I propose we work on that but it should not block this cleanup. Cleanup or not, the documentation still does not exists, is still up to the user to generate, and so the idea of using ansible-galaxy is good :)

eroussy: Seems good to me.
We should open an issue to upload the plugin in ansible galaxy, and remove the `module_documentation` part on the README.

 @dupremathieu are you okay with that ?

eroussy (changes requested):
Also, can you update the commit message to explain why you remove these directory ?
It will be better to undestand it in the future.

PR #562, opened 2024-07-15
deploy VM standalone + cluster: change tmp folder

Gabin-rte:
In case the qcow2 file is to big for /tmp
Brings back "ansible_remote_tmp" used to specify where ansible does tmp copy (very usefull for large vms)


playbooks/deploy_vms_standalone.yaml
```diff
@@ -29,10 +29,20 @@
...
+        owner: "{{ ansible_user }}"
+        group: "{{ ansible_user }}"
+        mode: '0755'
+      when: qcow2tmpuploadfolder is defined
     - name: Copy the disk on target
       copy:
         src: "{{ hostvars[item].vm_disk }}"
         dest: "{{ disk_pool }}/{{ hostvars[item].inventory_hostname }}.qcow2"
+      vars:
+        ansible_remote_tmp: "{{ qcow2tmpuploadfolder | default(omit) }}"
```
eroussy, line 45: It seems to me that the variable `ansible_remote_tmp` should not be defined in the playbook, but either in the inventory or directly in ansible.cfg.
Otherwise, looks good to me.

dupremathieu: No, it is an exception here because qcow2 disk cannot fit inside tmp.

dupremathieu (approved):

dupremathieu (changes requested):
In commit message

Add an extra new line between the title and the body.

to big -> too big.

PR #565, opened 2024-07-16
Multiple changes in cluster_setup_network and prerequisdebian

Gabin-rte:
- Remove br0 from VMs: The br0 bridge is not needed in VMs. It is only useful for the hypervisor to share a NIC.
- Remove Build hosts file for standalone_machine and VMs: The hosts file is useful for cluster machines (ceph, pacemaker). If needed in standalone machines and VMs, the Push hosts file is the appropriate method.
- Push hosts file is now available for VMs.
- haclient group is only in cluster_machines: Moved out from physical_machines.


playbooks/cluster_setup_network.yaml
```diff
@@ -58,8 +58,6 @@
   become: true
   hosts:
     - VMs
-  vars_files:
-    - ../vars/network_vars.yml
```
eroussy, line 62: if you remove this `vars_files`, you can remove this entire section "Apply network with systemd-networkd roles on VMs" (line 57-67).
But the admin IP address will not be set in the VM, we will have to define the .network file in the VM inventory.

eroussy (changes requested):
Can you please split this commit into the different features you propose ?

And also reduce the commit lines to 72 characters.
If you seek for the complete seapath contributing guidelines, they are listed here : https://github.com/seapath/.github/blob/main/CONTRIBUTING.md#git-workflow-guidelines

playbooks/cluster_setup_network.yaml
Gabin-rte: In order to play this task you need to access the VM. Thus admin IP is already configured (maybe with SEAPATH build_iso project) in VM disk and changing this configuration with a "br0" bridge is not something you need. However you might want to update the configuration or configure others NIC. To do so systemd_networkd role is the way to go, with variables like custom_network, systemd_networkd_link and custom_netdev.

eroussy: agree, but  `custom_network` and `custom_netdev` are configured and defined in this file `vars/network_vars.yml` so if you remove it, you can't use them anymore.

I think the best would be to create two new variables, may be `seapath_network` and `seapath_netdev`  that contains the entire network and netdev definition of the VM and that are passed to systemd_networkd role.
This would may be require a new vars file, but it would be way simpler than the previous one.

PR #567, opened 2024-07-17
Changes in cluster_setup_network and prerequisdebian + Refactor playbook deploy_vms_standalone

Gabin-rte:
- Build the hosts file, and use ansible_host as a fallback if ip_addr is not defined.
- Push hosts file is now available for VMs.
- haclient group is only in cluster_machines: Moved out from physical_machines.
- Refactor playbook deploy_vms_standalone to use loop and fix variable usage
  - Changed with_items to loop for consistency and modern Ansible practices
  - Use force variable to upload an already running VM as in deploy_vms_cluster
  - Use enable variable to start a VM as in deploy_vms_cluster

### DIFF
--- a/playbooks/cluster_setup_network.yaml
+++ b/playbooks/cluster_setup_network.yaml
@@ -260,6 +260,7 @@
   hosts:
     - cluster_machines
     - standalone_machine
+    - VMs
   become: true
   tasks:
     - block:
--- a/playbooks/cluster_setup_network.yaml
+++ b/playbooks/cluster_setup_network.yaml
@@ -248,7 +248,7 @@
     - name: Build hosts file
       lineinfile:
         dest: /etc/hosts
-        line: "{{ ip_addr }} {{ inventory_hostname }} {{ hostname | default('') }}"
+        line: "{{ ip_addr | default( ansible_host ) }} {{ inventory_hostname }} {{ hostname | default('') }}"
         state: present
     - name: remove 'debian' from hosts file
       lineinfile:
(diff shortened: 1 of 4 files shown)


playbooks/deploy_vms_standalone.yaml
```diff
@@ -37,54 +47,75 @@
...
     - name: Create VMs
       virt:
         command: define
         xml: >-
           {{ lookup('template',
           hostvars[item].vm_template,
           template_vars=dict(vm=hostvars[item])) }}
-      with_items: "{{ groups['VMs'] }}"
+      loop: "{{ groups['VMs'] }}"
+      when: item not in all_vms.list_vms or (item in all_vms.list_vms and hostvars[item].force is defined and hostvars[item].force)
```
eroussy, line 111: I would be clearer to create an Ansible block with this the previous tasks (L51-L111)

eroussy (changes requested):

playbooks/deploy_vms_standalone.yaml
Gabin-rte: It is not possible to loop on a block. Solution is to create another file.

eroussy: Oh, you are right, my bad.
Ok to leave it as it is.

playbooks/cluster_setup_network.yaml
```diff
@@ -58,8 +58,6 @@
   become: true
   hosts:
     - VMs
-  vars_files:
-    - ../vars/network_vars.yml
```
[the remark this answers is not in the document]
Gabin-rte: Change dropped.

PR #569, opened 2024-07-22
ci-yocto-weekly: increase pcap loop

bchedotel:
Now that we've corrected the problem with the long test script, we can increase the number of pcap loops,which will increase the test duration.


dupremathieu:
@eroussy Do we need to keep this PR?

eroussy:
The CI will be refactored to be used on all machines. This will be handled at this moment.

PR #585, opened 2024-08-20
Apply tuned profile to VM

### LINKED ISSUE #529: Apply tuned profile on SEAPATH virtual machines
Tuned is used to configure the real-time capabilities in the Debian kernel.
A custom Seapath tuned profile is applied to the hypervisor in https://github.com/seapath/ansible/blob/f087cef8a740638689c68ae124d27e64cac63d6a/roles/debian/hypervisor/tasks/main.yml#L167
A tuned profile must also be applied on SEAPATH VM. A simple `realtime-virtual-guest` could do the job

eroussy:
Apply realtime-virtual-guest inside the Debian virtual machines. When not applied, hours-long cyclictest have a maximum latency over 10ms.

This commit also creates a new debian/vm roles for this purpose.

Closes #529

### DIFF
--- a/playbooks/ci_prepare_VMs.yaml
+++ b/playbooks/ci_prepare_VMs.yaml
@@ -22,3 +22,18 @@
         name: systemd-resolved
         state: stopped
         enabled: false
+
+    - name: remove tuned dynamic tuning
+      lineinfile:
+        path: /etc/tuned/tuned-main.conf
+        state: present
+        regexp: '^#?dynamic_tuning = .*$'
+        line: 'dynamic_tuning = 0'
+    - name: enable tuned.service
+      ansible.builtin.systemd:
+        name: tuned.service
+        enabled: yes
+        state: restarted
+    - name: load realtime-virtual-guest tuned profile
+      ansible.builtin.command: tuned-adm profile realtime-virtual-guest
+      when: "'rt' in vm_features"


insatomcat:
I don't think Seapath should be configuring the VMs, at all. Seapath uses VMs to be able to provide a multi-vendor system (in a monovendor system, containers should be used).
The VM is the responsability of the (VM) vendor. Seapath must have the required feature to manage the VM (deployment, administration), but configuring it in not in scope.
I know we use test VMs for the CI and benchmarking, and that is very relevant, however in that case we provide (as any vendor should) the VM already configured (xml + qcow2), there is no need for seapath playbooks to go into the VMs to configure it. It that case we can the seapath project as a vendor for a "test VM", but this is very unique and specific, and does not really "need" post deployment configuration (it should be preconfigured).

I don't see any real world use case for seapath to go into VM configuration.

eroussy:
I agree with you, the main goal of SEAPATH is to deployed third parties VM, however : 
1. We use test VM in the CI and we should launch cyclictest in it IMO.
2. This provide an example of how to configure a Debian VM
3. I know that some users uses SEAPATH VM as a base to deploy their applications in it.
So I still think that this is relevant

Also, SEAPATH ansible already modify the VMs (network + prerequisites + hardening) so this commit only follows this trend.
May be we should refactor this, in order to distinguish between
- The VMs to deploy on SEAPATH (could be SEAPATH VM or third parties)
- The VMs on which to apply the SEAPATH playbooks (network + prerequisites + hardening)
I will not have the time to do it now, but i can open an issue about that.

eroussy:
So what we could do is : 
1. Move this tuned profile inside a CI playbook (to integrate cyclictest in CI later)
2. Begin to remove all VMs mentions in the standard playbooks (for example cluster_setup_network.yaml)
Does it seems right to you ?

insatomcat:
It seems great.

dupremathieu (approved):

PR #590, opened 2024-08-23
Refactor entry playbook debiancentos

eroussy:
This PR configures the playbook `seapath_setup_main.yaml` to be the entry point of all SEAPATH configurations (standalone or cluster / Debian or CentOS)
It also rename some "cluster" playbooks that were applied also on standalone machines and update the README.
This PR is in the process of merging debiancentos and main branches (See equivalent PR on main branch #589 )

API changes:
- `playbooks/cluster_setup_debian.yaml` is removed. Use `playbooks/seapath_setup_main.yaml` instead.
- `playbooks/cluster_setup_centos.yaml` is removed. Use `playbooks/seapath_setup_main.yaml` instead.
- `playbooks/cluster_setup_all.yaml` is renamed to `playbooks/seapath_setup_main.yaml`
- `playbooks/cluster_setup_network.yaml` is renamed to `playbooks/seapath_setup_network.yaml`
- `playbooks/cluster_setup_prerequisdebian.yaml` is renamed to `playbooks/seapath_setup_prerequisdebian.yaml`
- `playbooks/cluster_setup_prerequiscentos.yaml` is renamed to `playbooks/seapath_setup_prerequiscentos.yaml`


yairpod:
Looks good to me

playbooks/cluster_setup_add_livemigration_user.yaml
```diff
@@ -7,7 +7,7 @@
 
 ---
 - name: Create livemigration user
-  hosts: hypervisors
+  hosts: hypervisors:&cluster_machines
```
insatomcat, line 10: standalone machines are hypervisors by nature.
I think hypervisors is a group that should only be a subgroup of cluster_machines (used just to make the distinction with observers).
That way the hosts syntax here could keep on being "hypervisors".
And for things that should apply to hypervisors and standalones we could use a more readable "OR" hosts syntax "hypervisors, standalone_machine" (instead of this "AND" syntax).

insatomcat (changes requested):
Other remarks : 
- do we really need 6 different commits for this one functionnal change?
- can we have the API changes writen in the commit instead of the PR description. We agreed on this, we need the important info (the why/how/impacts) to be written in the commits

playbooks/cluster_setup_add_livemigration_user.yaml
eroussy: The syntax is not pretty, I agree. But I like the fact that "cluster machine" is mentioned.
Also, what you propose would require to have a different inventory structure between standalone and cluster machines, I don't know if it is a good idea.

In both cases, I am mostly focused on merging the branches, so I propose to open an issue (or a slack thread) to discuss with that and handle it in another PR.

insatomcat: By the way, it seems it's already the accepted logic to use an OR filter for hypervisors and standalones:
https://github.com/seapath/ansible/blob/f7ef1b1886143b998d07e1640eb47a4fae6d06cd/playbooks/seapath_setup_prerequiscentos.yaml#L22
https://github.com/seapath/ansible/blob/f7ef1b1886143b998d07e1640eb47a4fae6d06cd/playbooks/seapath_setup_prerequisdebian.yaml#L22

eroussy: IMO: the files you send are wrong and should be changed to "hypervisors" only.
It is important to get the information on the playbooks if the files are targeted on cluster hypervisors or standalone hypervisors

eroussy: > then shouldn't this PR be targeted on merged-main?

merge-main will only be used to do the standalone POC.
It is important to first get the two branches as close as possible.
It is a matter of dividing the work, if you prefer.

insatomcat: we could rename "hypervisors" and "observers" to "cluster_hypervisors" and "cluster_observers". And the standalone machines would stop being "hypervisors" and be only "standalone_machine".

That would show the "cluster" word you like so see, and avoid weird hosts syntax.
What do you think?

dupremathieu: I agree that it will be great to avoid the negation.

The observer machine could just be named observer, there is no need for to put it in a group (SEAPATH does not support multiple observers).

There is also the Ansible tag feature we can use to know whether a machine is a cluster machine or not.

eroussy: So what we do is :
- removing the `hypervisors` group
- putting the `observer` machine outside of the `cluster_machines` group

Does it seems right to you ?

( Anyway, this PR is related to the merge of the playbook, so I will do it to another PR just after the one I already opened)

insatomcat: That's not what I had in mind.
A host can be:
 - either a standalone machine (in which case we now it's an hypervisor, we don't need to put it in a group for that)
 - a cluster_machine
    - if it's a cluster_machine, it can be an hypervisor or a observer
    - but since "hypervisor" can be undertood as including the standalone machine I suggest we prefix the names and so have "cluster_hypervisor" and "cluster_observers" or just "observers" (I don't really care that only one is supported for now, it's a group so plural anyway).

that way 
- all the "hypervisor" specific roles are applied to "cluster_hypervisors" and "standalone_machines" (OR syntax)
- all the "cluster node" roles are applied to "cluster_machines" group
- all the "standalone" roles are applied to "standalone_machines" group
- the specific "livemigration" roles are applied to "cluster_hypervisors" group

Does that make sense?

eroussy: Both looks good to me.

I kickstart a PR tomorrow or the day after. Let's discuss it on this PR.

insatomcat (dismissed):

dupremathieu (approved):

PR #593, opened 2024-08-26
Harmonize network playbook debiancentos

eroussy:
Refactor the network playbooks to be closer to Yocto one.
Also put some tasks out of the playbook cause they have nothing to do with network configuration.
See the equivalent PR on main branch : #592 

API changes:
- PTP/NTP synchronization is now configured in playbooks/seapath_setup_time_synchronization.yaml
- snmp role is called from the main playbook instead of from the network playbook
- network inside VMs is not configured anymore. The user should write its own playbook if he want to do that.

There is no changes for the user if he uses the seapath_setup_main.yaml playbook.


eroussy:
This work will be integrated in another pull request that moves timemaster to a role.

PR #594, opened 2024-08-27
fix virt-df.sh real location

insatomcat:
This script allows the host to know the available disk space of guests, so that it can expose it with snmp


src/debian/snmp/virt-df.sh
```diff
@@ -1,49 +1,76 @@
 #!/bin/bash
+
```
eroussy, line 2: Please add a small description of what this script is doing.
Also, it needs copyright and license.

eroussy (changes requested):

src/debian/snmp/virt-df.sh
insatomcat: ok

eroussy (approved):

PR #596, opened 2024-08-27
add qDevice and qNetd configuration

dupremathieu:
This PR adds the configuration of qDevice and qNetd daemons to the cluster_setup_ha playbook. The qDevice and qNetd daemons are used to provide a quorum without the need of a third corosync node. This is useful when the cluster has only two or an even number of nodes.

### DIFF
--- a/ansible-requirements.yaml
+++ b/ansible-requirements.yaml
@@ -12,7 +12,7 @@ roles:
     - name: "corosync"
       src: "https://github.com/seapath/corosync-ansible"
       scm: git
-      version: "d7e30a7ad979e08af5e9a5696deb8039e6f52d78"
+      version: "5a99ca31b82b5d416ce3ee21f251411b4f461d5c"
 
 collections:
     - name: "community.libvirt"
--- a/playbooks/cluster_setup_ha.yaml
+++ b/playbooks/cluster_setup_ha.yaml
@@ -6,8 +6,21 @@
 # also be called independently.
 
 ---
-- name: Configure SEAPATH specific files
+- name: Search for corosync nodes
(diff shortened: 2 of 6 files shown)


playbooks/cluster_setup_ha.yaml
```diff
@@ -79,27 +95,196 @@
...
             mode: '0400'
-        - name: Start pacemaker
+        - name: Start corosync
           ansible.builtin.systemd:
               name: corosync
               state: started
               enabled: true
       when: groups['valid_machine'] is defined
 
+# Configure qDevice and qNetd
```
Revalioli, line 119: Could this be put in a sub-playbook or in roles ? This is a lot of tasks, and the file would be >300 lines long

Revalioli: Also it would allow having one check on the `corosync_use_qdevice` variable

ansible-requirements.yaml
```diff
@@ -12,7 +12,7 @@ roles:
     - name: "corosync"
       src: "https://github.com/seapath/corosync-ansible"
       scm: git
-      version: "d7e30a7ad979e08af5e9a5696deb8039e6f52d78"
+      version: "5a99ca31b82b5d416ce3ee21f251411b4f461d5c"
```
Revalioli, line 15: In the final PR, merge first the PR in corosync-ansible and then simply state that you update the ref in the commit message

Revalioli (changes requested):
See comments

dupremathieu:
qDevice / qNetd are too complex.

PR #600, opened 2024-08-30
move everything into autonomous and independent roles

insatomcat:
Massive refactor of playbooks logic to autonomous and independent roles.
This makes Seapath very modular, and should help with the merge. This also allows to publish the majority of the seapath code in an ansible collection on galaxy.

At the end of the whole process, there is no file left in the ./src/ and ./templates/ folder, since all files and templates are now in there respective role folders.

This is just code organizing, there is no API change.


dupremathieu (changes requested):
Change the "test" commit message otherwise LGTM.

insatomcat:
> Change the "test" commit message otherwise LGTM.

That was a test commit, I just dropped it and rebased everything on debiancentos

dupremathieu (approved):

PR #612, opened 2024-09-09
Add snmp v3 support

insatomcat:
This create the snmp v3 user accounts.
API Change : new inventory variable for the snmp v3 user accounts:
```
 snmp_accounts:
    - name: user1
      password: mypassword1
    - name: user2
      password: mypassword2
    - name: user3
      password: mypassword3
```

### DIFF
--- a/inventories/seapath_cluster_definition_example.yml
+++ b/inventories/seapath_cluster_definition_example.yml
@@ -358,3 +358,12 @@ all:
           IPv4_address 127.0.0.0/8
           IPv4_address 172.0.0.0/8
           IPv4_address 192.0.0.0/8
+
+        # Account to be created for SNMP v3 support. Optional.
+        #snmp_accounts:
+        #  - name: user1
+        #    password: mypassword1
+        #  - name: user2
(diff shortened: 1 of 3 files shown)


inventories/seapath_cluster_definition_example.yml
```diff
@@ -358,3 +358,9 @@ all:
           IPv4_address 127.0.0.0/8
           IPv4_address 172.0.0.0/8
           IPv4_address 192.0.0.0/8
+
+        # Account to be created for SNMP v3 support. Optional.
+        #snmp_accounts:
+        #  - "createUser user1 SHA mypassword1 AES mypassword1"
```
eroussy, line 364: These are technically not "accounts", but commands to create accounts.
Would it be possible to add only the username and the password in the inventory, and make the playbook create the command ?

insatomcat: MD5 and DES are depreciated, so we'll impose SHA and AES.

dupremathieu (approved):

PR #613, opened 2024-09-10
Add VM start at boot for standalone

Gabin-rte:


eroussy (commented):
It seems to me that what you want to achieve is the same result as the command `virsh autostart vm_name`. It would be better to launch this command after VM deployment, instead of doing a service this way.

Unfortunately, it seems that we cannot write this behavior directly in the xml

Gabin-rte:
> It seems to me that what you want to achieve is the same result as the command `virsh autostart vm_name`. It would be better to launch this command after VM deployment, instead of doing a service this way.
> 
> Unfortunately, it seems that we cannot write this behavior directly in the xml

It does not work as expected. For exemple, in case of power lost VMs will not start at boot. 
Maybe it’s a bug in my platforms, can you confirm this behavior?

dupremathieu (changes requested):
Remove the merge commit (use rebase).

Why do you not use libvirt autostart flag (e.g. virsh autostart my_vm)?

dupremathieu:
> > It seems to me that what you want to achieve is the same result as the command `virsh autostart vm_name`. It would be better to launch this command after VM deployment, instead of doing a service this way.
> > Unfortunately, it seems that we cannot write this behavior directly in the xml
> 
> It does not work as expected. For exemple, in case of power lost VMs will not start at boot. Maybe it’s a bug in my platforms, can you confirm this behavior?

On my side, with autostart, VMs start even in case of power lost.

insatomcat:
solved by #614

PR #615, opened 2024-09-16
Move Update grub to variable for CentOS support

yairpod:
CentOS uses a different grub update command then the Debian.
The CI playbooks need to be updated.


dupremathieu (approved):

roles/ci_restore_snapshot/vars/Seapath_host_Yocto_distribution.yml
```diff
@@ -0,0 +1,5 @@
+# Copyright (C) 2024, RTE
+# SPDX-License-Identifier: Apache-2.0
+
+---
+grub_update_command: "update-grub"
```
eroussy, line 5: This is not useful, we will not use the snapshot restore in Yocto CI.

Otherwise, LGTM

eroussy (approved):

PR #616, opened 2024-09-17
pacemaker_ra/VirtualDomain: add support of sync_config_on_stop

dupremathieu:
This patch adds the support of the "sync_config_on_stop parameter" when "seapath" parameter is set to "true".

When the "sync_config_on_stop" parameter is set to "true", the domain configuration is saved to the rbd image metadata when the resource is stopped.

The sync config validates the XML configuration and requires libxml2-utils packages installed by this PR: https://github.com/seapath/build_debian_iso/pull/117.


insatomcat:
I guess it does not hurt to make this option work with ceph rbd if someone really wants it, but I for one would never use this.
Dumpxml dumps a lot of data (all the hidden defaults, all the pci allocations, etc.) that I don't want to see in the VM xml. I like to have this xml clean and readable, easier to amend.
Seapath comes with the edit_metadata.sh tool where you can easily change the VM configuration and restart the VM to apply the change.

insatomcat (dismissed):

PR #617, opened 2024-09-24
OVS management through Netplan and Libvirt

Paullgk:
This PR adds new ways to managed OVS bridges and virtual port through Netplan and Libvirt, on a standalone Seapath machine.
* Using Netplan configuration file [inventories/netplan_ovs_bridge_example.yaml.j2](https://github.com/seapath/ansible/compare/debiancentos...netplan_ovs?expand=1#diff-4ddaad927cd1722f5584e18d4e721c6983495a0cc458ee47d77712c3170e040b), it is now possible to create an OVS bridge from Netplan.
* Using Libvirt <virtualport> field, it is now possible to create OVS virtual port directly from Libvirt XML configuration file. The modification adds also the possibility to configure a VLAN. 
These modifications are based on the current inventory OVS bridge  syntax.

### DIFF
--- a/inventories/seapath_vm_definition_example.yml
+++ b/inventories/seapath_vm_definition_example.yml
@@ -15,8 +15,10 @@ all:
           preferred_host: pc1 # Optional
           force: true
           bridges:
-            - name: "br0" # Change the bridge name
+            - name: "ovs0" # Change the bridge name
               mac_address: "52:54:00:c4:ff:02" # change the MAC address
+              vlan:
+                vlan: 100 # Add a VLAN on the OVS port created
         guest1:
           description: "Complex VM"
           ansible_host: 192.168.212.131 # Update the VM IP
--- a/templates/vm/guest.xml.j2
+++ b/templates/vm/guest.xml.j2
@@ -145,6 +145,11 @@
             {% if bridge.type is defined %}
                 <virtualport type='{{ bridge.type }}'/>
             {% endif %}
+            {% if bridge.vlan is defined %}
+                <vlan>
+                    <tag id='{{ bridge.vlan.vlan_tag }}'/>
+                </vlan>
+            {% endif %}
         </interface>
         {% endfor %}
         {% endif %}
--- a/templates/vm/guest.xml.j2
+++ b/templates/vm/guest.xml.j2
@@ -142,6 +142,9 @@
             <source bridge="{{ bridge.name }}"/>
             <mac address="{{ bridge.mac_address }}"/>
             <model type="virtio"/>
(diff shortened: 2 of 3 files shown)


templates/vm/guest.xml.j2
```diff
@@ -142,6 +142,13 @@
             <source bridge="{{ bridge.name }}"/>
             <mac address="{{ bridge.mac_address }}"/>
             <model type="virtio"/>
+            <virtualport type='openvswitch'/>
+            <address type='pci' domain='0x0000' bus='0x03' slot='0x01' function='0x0'/>
```
insatomcat, line 146: is it a good idea to fix the pci allocation details?

inventories/seapath_vm_definition_example.yml
```diff
@@ -17,6 +17,8 @@ all:
           bridges:
             - name: "br0" # Change the bridge name
               mac_address: "52:54:00:c4:ff:02" # change the MAC address
+              vlan:
+                vlan: 100 # Add a VLAN on the OVS port created
```
insatomcat, line 21: br0 is not an OVS bridge (in your example it's "ovs0"), so maybe this example is misleading

insatomcat (changes requested):

insatomcat:
Can you merge in one commit? This does not justify 4 differents commits...

templates/vm/guest.xml.j2
Paullgk: Nope, I removed them

inventories/seapath_vm_definition_example.yml
Paullgk: Fixed

Paullgk:
> Can you merge in one commit? This does not justify 4 differents commits...

As the modification are on 3 different files I would prefer to keep this granularity, easier in case of revert and for commit reliability

insatomcat:
I find it easier to revert the whole change, and easier to understand the change if its explanation is not split in different commit descriptions. We had a talk about this in maintainers meeting.
1 feature or 1 fix = 1 commit with everything explained in its description (we do not count on the PR description since it not written in git).
@eroussy do you agree?

dupremathieu:
> 1 feature or 1 fix = 1 commit with everything explained in its description (we do not count on the PR description since it not written in git). @eroussy do you agree?

Here there are more than 1 feature, no?

insatomcat:
if you feel there are different features then multiple commits are fine. I felt that justifying 4 full features with 25 lines changed in 3 files was a bit of a stretch but hey why not.

insatomcat (approved):

PR #621, opened 2024-10-02
ceph-ansible: move to ceph 16

dupremathieu:
The Ceph and ceph-ansible have to be in the same version. This patch moves ceph-ansible to the ceph 16 version to match the ceph version.

This patch upadtes the ceph-ansible submodule, the ansible-requirements.yaml file to match the new ceph-ansible requirements and update the patches.

### DIFF
--- a/.cqfd/docker/Dockerfile
+++ b/.cqfd/docker/Dockerfile
@@ -22,7 +22,15 @@ RUN set -x \
         vim \
     && sed -i '/en_US.UTF-8/s/^# //g' /etc/locale.gen && locale-gen \
     && python3 -m venv --system-site-packages /env \
-    && /env/bin/pip install ansible==2.10.7 ansible-lint==5.4.0 yamllint six netaddr jmespath \
+    && /env/bin/pip install \
+        ansible==2.10.7 \
+        ansible-lint==5.4.0 \
+        yamllint \
+        six \
+        netaddr \
+        jmespath \
+    && cp /env/lib/python3.11/site-packages/pip/_vendor/six.py \
+        /env/lib/python3.11/site-packages/ansible/module_utils/six/__init__.py \
     && rm -rf /var/lib/{apt,dpkg,cache,log}/*
 
 ENV PATH="/env/bin:/env/usr/bin:/env/sbin:/env/usr/sbin:$PATH"
--- a/.cqfd/docker/Dockerfile
+++ b/.cqfd/docker/Dockerfile
@@ -1,42 +1,30 @@
 # Copyright (C) 2021, RTE (http://www.rte-france.com)
 # SPDX-License-Identifier: Apache-2.0
 
-FROM ubuntu:22.04
+FROM debian:12
 
-ENV DEBIAN_FRONTEND noninteractive
-
-ENV LANG en_US.UTF-8
-ENV LANGUAGE en_US.UTF-8
-ENV LC_ALL en_US.UTF-8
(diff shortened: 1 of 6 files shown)


.cqfd/docker/Dockerfile
```diff
@@ -36,7 +37,6 @@ RUN set -x \
     && sed -i "s/# en_US\.UTF-8 UTF-8/en_US\.UTF-8 UTF-8/" /etc/locale.gen \
     && locale-gen \
     && dpkg-reconfigure locales \
-    && pip install "ansible>=2.9,<2.10,!=2.9.10" \
```
Paullgk, line 39: Ceph-ansible 6.0 only supports Ansible 2.10: https://docs.ceph.com/projects/ceph-ansible/en/latest/#releases
We need to keep the Ansible version check.

Paullgk (changes requested):

.cqfd/docker/Dockerfile
dupremathieu: The apt version is Ansible 2.10.
I have removed the pip install and replaced it by apt version.

insatomcat: why not use directly the Dockerfile we use in debiancentos? if the idea is to merge in the end...
https://github.com/seapath/ansible/blob/debiancentos/.cqfd/docker/Dockerfile

dupremathieu: Good point, I have made the change.

.cqfd/docker/Dockerfile
```diff
@@ -12,6 +12,7 @@ ENV LC_ALL en_US.UTF-8
 RUN set -x \
     && apt-get update \
     && apt-get install -y \
+        ansible \
```
Revalioli: Why don't you squash commit https://github.com/seapath/ansible/pull/621/commits/6a448643c220682222330d4283bedd83cd6d90a9 in this one ? It would be more clear in the git history

If you do so, remember to update the commit message.

Revalioli (changes requested):
Squash 6a448643c220682222330d4283bedd83cd6d90a9 into 58010dbf47365d2f3717f0886d8772dbc32f9e07

.cqfd/docker/Dockerfile
Paullgk: I do not agree with this: we are not sure that Ansible will be packaged in 2.10 version in 22.04. Using Pip, we can specify the version we want. You should remove these lines.

.cqfd/docker/Dockerfile
Paullgk, line 39: Same as my remark above

Paullgk (changes requested):

.cqfd/docker/Dockerfile
dupremathieu: Because I have forgotten to update the branches.

I no longer use these commits.

Paullgk (changes requested):
You removed my commit which bumped the version of python3-six used by the Ansible payload. Pay attention that without it, Ansible 2.10 will break as soon we will be using Python 3.12 (Yocto Scarthgap)

dupremathieu:
> You removed my commit which bumped the version of python3-six used by the Ansible payload. Pay attention that without it, Ansible 2.10 will break as soon we will be using Python 3.12 (Yocto Scarthgap)

I have re-applied it with adaptation to work the debiancentos Dockerfile.

Paullgk (approved):

PR #623, opened 2024-10-05
HA playbook: prevent automatic restart of corosync

insatomcat:
This commit only restarts corosync when needed (the corosync.conf file has changed).

### DIFF
--- a/playbooks/cluster_setup_ha.yaml
+++ b/playbooks/cluster_setup_ha.yaml
@@ -61,11 +61,19 @@
         src: ../src/debian/corosync.conf.j2
         dest: /etc/corosync/corosync.conf
       when: groups['valid_machine'] is undefined
-    - name: Making sure that Corosync service is started
+      register: corosync_conf
+    - name: Restart Corosync
       ansible.builtin.systemd:
         name: corosync
         state: restarted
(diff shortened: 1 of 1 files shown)


playbooks/cluster_setup_ha.yaml
```diff
@@ -61,11 +61,19 @@
         src: ../src/debian/corosync.conf.j2
         dest: /etc/corosync/corosync.conf
       when: groups['valid_machine'] is undefined
-    - name: Making sure that Corosync service is started
+      register: corosync_conf
```
eroussy, line 64: You should use a handler to do that, IMO

insatomcat: I would agree because right now this playbook is made of several shorts plays, and so the handler would be run just after this task.
But my target with #600 is to have this whole logic in a role. Since the handler runs after the play, the service would only be restarted after the whole role is run, but I need the service restarted sooner because there is still a lot to be done after this task.
So If I move this task to a handler, I would have to create a new handler file, and then add a flush_handlers to run it sooner... This works but seems more complicated (and so less maintanable) than just running a restarted task in the role...

eroussy (approved):

PR #626, opened 2024-10-18
library/cluster_vm: update to support new vm_manager options

dupremathieu:
Add support for:
- pacemaker_meta
- pacemaker_params
- pacemaker_utilization
- clear_pacemaker_meta
- clear_pacemaker_params
- clear_pacemaker_utilization
- disk_bus

### DIFF
--- a/library/cluster_vm.py
+++ b/library/cluster_vm.py
@@ -165,6 +165,47 @@ options:
         C(clone)
       - Metadata key must be composed of letters and numbers only
     type: dict
+  pacemaker_meta:
+    description:
+      - Pacemaker "meta" parameter to add
+      - This parameter is relevant only if the I(command) is C(create) or
+        C(clone), optional
+    type: dict
+  pacemaker_params:
+    description:
+      - Pacemaker "params" parameter to add
+      - This parameter is relevant only if the I(command) is C(create) or
+        C(clone), optional
+    type: dict
+  pacemaker_utilization:
+    description:
+      - Pacemaker "utilization" parameter to add
+      - This parameter is relevant only if the I(command) is C(create) or
+        C(clone), optional
+    type: dict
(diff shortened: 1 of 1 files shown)


library/cluster_vm.py
```diff
@@ -18,7 +18,7 @@
 
 short_description: Manages virtual machines on a SEAPATH cluster
 
-version_added: "2.9"
+version_added: "2.10"
```
Revalioli, line 21: According to [ansible documentation](https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html), this version field should not be changed
Even if compatible with Ansible 2.10, the module was still added with Ansible 2.9

library/cluster_vm.py
```diff
@@ -165,6 +165,44 @@
         C(clone)
       - Metadata key must be composed of letters and numbers only
     type: dict
+  pacemaker_meta:
+    description:
+      - Pacemaker "meta" parameter to add
+      - This parameter is optional if the I(command) is C(create) or C(clone)
```
Revalioli, line 171: State that it is **only used with create or clone command** and that it is optional.
Otherwise this message gives the impression it can be used for other commands.
Same remark for below help strings

library/cluster_vm.py
```diff
@@ -304,6 +351,7 @@
     src_name: guest0
     command: clone
     xml: "{{ lookup('template', 'my_vm_config.xml', errors='strict') }}"
+    clear_pacemaker_params: False
```
Revalioli, line 354: False is already the default value, maybe replace by "true" for the example

Revalioli (changes requested):
Some remarks on documentation

Revalioli (approved):

PR #632, opened 2024-11-05
ssc600.xml.j2: create a template for the ABB VMs

Paullgk:
The ssc600 SW needs to be configured through a libvirt xml file. It can't be created using the default guest.xml.j2 template from Seapath due to his slight differences. Thus, the ssc600.xml.j2 is created specially for the ssc600 VMs.


eroussy:
This will be done in another PR

PR #633, opened 2024-11-05
ssc600.xml.j2: create a template for the ABB VMs

Paullgk:
The ssc600 SW needs to be configured through a libvirt xml file. It can't be created using the default guest.xml.j2 template from Seapath due to his slight differences. Thus, the ssc600.xml.j2 is created specially for the ssc600 VMs.

### DIFF
--- a/templates/vm/ssc600.xml.j2
+++ b/templates/vm/ssc600.xml.j2
@@ -1,0 +1,118 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--  Copyright (C) 2024 Savoir-faire Linux, Inc-->
+<!-- SPDX-License-Identifier: Apache-2.0 -->
+
+{% set cpu_nb = (vm.cpuset | length) if vm.cpuset is defined else vm.nb_cpu | default(1) %}
+
+<domain type="kvm">
+    <name>{{ vm.inventory_hostname }}</name>
+    <uuid>{{ vm.uuid |default( 9999999999999999999999 | random | to_uuid) }}</uuid>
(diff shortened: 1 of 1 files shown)


templates/vm/ssc600.xml.j2
```diff
@@ -0,0 +1,118 @@
...
+           <target dev="vda" bus="virtio"/>
+        </disk>
+	{% endfor %}
+	{% endif %}
+    <controller type="usb" index="0" model="none"/>
+    <controller type="pci" index="0" model="pcie-root"/>
+    <controller type="virtio-serial" index="0"/>
+    <filesystem type="mount" accessmode="mapped">
+        <driver type="path" wrpolicy="immediate"/>
+        <source dir="/var/lib/libvirt/images/ptp"/>
```
dupremathieu, line 79: Maybe we should add a script hook to create this directory and export the ptp status.

insatomcat:
the deploy_vms_cluster playbook allows 3 methods to provide the xml file:
- either you provide the xml file directly with the xml_path variable
- or you provide the template to use
- or an xml file named after the VM is provided in the vms_disks_directory folder

https://github.com/seapath/ansible/blob/12b59211dd58ed81160315b48fc441e708ae3013/roles/deploy_vms_cluster/tasks/main.yml#L47

Your problem can easily be solved in the cluster deployment with either providing the file in the vms_disks_directory or via the xml_path variable.
Maybe you should port this behavior to the standalone deployment role.

eroussy:
This will be done in another PR

PR #639, opened 2024-11-22
do not use netplan to create team0

dupremathieu:
To avoid a bug on Netplan during OVS cleanup step that reset the RSTP configuration of the bridge that is already configured, we decided not to use Netplan to create the team0 interface.

The team0 bridge is now created in anycase by the systemd-networkd.


eroussy (commented):
Commit message : 
To avoid a bug **on** Netplan **during** OVS cleanup step that reset**s**

vars/network_team_only_vars.yml
```diff
@@ -0,0 +1,29 @@
+# Copyright (C) 2021, RTE (http://www.rte-france.com)
```
eroussy, line 1: Is it not possible to have : 
- One systemd-network var file for the team0 bridge
- One systemd-network var file for the rest of the network
The first file would always be included, and the second one only if netplan is not used.

In general, I think it would be better to have multiple systemd_networkd files (admin, cluster, ptp) as it is done for netplan.

eroussy (changes requested):

vars/network_team_only_vars.yml
dupremathieu: It is impossible to do it in this way.

I have changed to have 3 files.
* One for team0 always included
* One for a network with only team0 and custom network, it relies on the variable defined in the team0 file.
* One for the default network (br0 and others). It also uses the team0 variable defined in the team0 file.

playbooks/seapath_setup_network.yaml
```diff
@@ -12,7 +12,11 @@
   become: true
   gather_facts: true
   vars_files:
-    - ../vars/network_vars.yml
+    - "../vars/network_team0.yml"
+  pre_tasks:
+    - name: Dynamically include network configuration variables
+      include_vars:
+        file: "../vars/network_{{ 'empty' if netplan_configurations is defined or (network_simple is defined and network_simple) else 'defaults' }}.yml"
```
eroussy, line 19: If you create this variable, you need to document it.
Add it to the inventories examples.

eroussy (changes requested):

eroussy (dismissed):

PR #643, opened 2024-11-27
Merge debiancentos yocto

TanguyRaufflet:
This PR adds commits to unify the main branch (Yocto) and the debiancentos branch.


playbooks/cluster_setup_ceph.yaml
```diff
@@ -18,17 +18,17 @@
   become: true
   gather_facts: yes
   roles:
-    - ceph_expansion_vg
-
+    - role: ceph_expansion_vg
+      when: ansible_distribution | regex_search("Debian")
```
dupremathieu, line 22: No it should work on Yocto too. Furthermore, it runs only when ansible_lvm is defined and lvm_volumes.

playbooks/cluster_setup_ceph.yaml
```diff
@@ -18,17 +18,17 @@
...
   hosts:
       osds
   become: true
   gather_facts: yes
   serial: 1
   roles:
-    - ceph_expansion_lv
-
+    - role: ceph_expansion_lv
+      when: ansible_distribution | regex_search("Debian")
```
dupremathieu, line 31: Some remarks.

roles/ceph_prepare_installation/tasks/main.yml
```diff
@@ -18,6 +18,10 @@
       state: directory
   register: new_ceph_installation
 
+- name: Cleanup Ceph OSD disks with ceph-volume
```
dupremathieu, line 21: The Prepare OSD disk should works on Yocto.

roles/add_livemigration_user/tasks/main.yml
```diff
@@ -8,8 +8,8 @@
       name: "{{ livemigration_user }}"
       shell: /bin/bash
       system: true
-      group: libvirt
-      create_home: true
+      groups: qemu,haclient,libvirt
+      create_home: false
```
dupremathieu, line 12: This will not work on Debian and Centos.

roles/add_livemigration_user/tasks/main.yml
```diff
@@ -24,32 +24,23 @@
       ssh_key_file: .ssh/id_rsa
       ssh_key_passphrase: ""
       force: false
-  - name: create livemigration .ssh directory
+  - name: create livemigration ssh directory
     file:
-      path: "/home/{{ livemigration_user }}/.ssh"
+      path: "/etc/ssh/{{ livemigration_user }}"
```
dupremathieu, line 29: Yocto specific.

roles/add_livemigration_user/tasks/main.yml
```diff
@@ -24,32 +24,23 @@
...
       state: present
-      path: "/home/{{ livemigration_user }}/.ssh/authorized_keys"
-      key: "{{ lookup('file','buffer/' + item + '-id_rsa.pub') }}"
-    with_items: "{{ groups['hypervisors'] }}"
-  - name: Copy the key to admin user's authorized_keys using Ansible module
-    authorized_key:
-      user: "{{ admin_user }}"
-      state: present
-      path: "/home/{{ admin_user }}/.ssh/authorized_keys"
+      path: "/etc/ssh/{{ livemigration_user }}/authorized_keys"
```
dupremathieu, line 43: Yocto specific.

dupremathieu (changes requested):

roles/add_livemigration_user/tasks/main.yml
TanguyRaufflet: Why can't we do the same as Yocto for Debian here ?

roles/add_livemigration_user/tasks/main.yml
TanguyRaufflet: Same question.

roles/add_livemigration_user/tasks/main.yml
TanguyRaufflet: because of the groups ?

dupremathieu: Maybe. But it the crate_home no which will be an issue during the ssh key setup.

roles/add_livemigration_user/tasks/main.yml
dupremathieu: `/etc/ssh/{{ livemigration_user }}/authorized_keys` is a Yocto specific path.

roles/add_livemigration_user/tasks/main.yml
```diff
@@ -3,13 +3,38 @@
...
+          create_home: true
+          ssh_path: "/home/{{ livemigration_user }}/.ssh"
+        CentOS:
+          groups: "libvirt"
+          create_home: true
+          ssh_path: "/home/{{ livemigration_user }}/.ssh"
+        Yocto:
+          groups: "qemu,haclient,libvirt"
+          create_home: false
+          ssh_path: "/etc/ssh/{{ livemigration_user }}"
```
insatomcat, line 30: This is not how we define distribution specific variable in a role.
You should add an `- include_vars: "{{ ansible_distribution | replace(' ', '_') }}.yml"` task at the beginning of the `tasks/main.yml file`, and then create the `CentOS.yml  Debian.yml  Seapath_host_Yocto_distribution.yml` files in the role's `vars` folder

roles/ceph_prepare_installation/tasks/main.yml
```diff
@@ -18,6 +18,10 @@
...
+  command: "ceph-volume lvm zap {{ ceph_osd_disk }} --destroy"
+  when: new_ceph_installation.changed and ansible_distribution | regex_search("Yocto")
```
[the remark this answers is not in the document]
insatomcat: same, why was this condition added?

roles/ceph_prepare_installation/tasks/main.yml
```diff
@@ -69,4 +73,4 @@
       when:
         - lvm_volumes is not defined or ansible_lvm['lvs'][lvm_volumes[0].data] is not defined
   when:
-    - new_ceph_installation.changed
+    - new_ceph_installation.changed and ansible_distribution | regex_search("Debian")
```
insatomcat, line 76: preparing the osd disk is not specific to Debian...

insatomcat (changes requested):

roles/add_livemigration_user/tasks/main.yml
TanguyRaufflet: Thanks !

roles/add_livemigration_user/tasks/main.yml
```diff
@@ -24,25 +27,26 @@
       ssh_key_file: .ssh/id_rsa
       ssh_key_passphrase: ""
       force: false
-  - name: create livemigration .ssh directory
+  - name: create livemigration ssh directory
     file:
-      path: "/home/{{ livemigration_user }}/.ssh"
+      path: "{{ ssh_path }}"
```
dupremathieu, line 32: In my opinion, we should rather change the ssh configuration on Yocto to use the regular one. The change was made due to the readonly feature before we added the /home directory in overlay.

dupremathieu (changes requested):

insatomcat:
The need to patch ceph-ansible is really specific to yocto, I'm not confortable changing ceph-ansible when the user is using centos or debian.
Could we find a way to make this patching optional, only for users of the yocto flavor?

TanguyRaufflet:
> ceph-ansible is really specific to yocto, I'm not confortable changing ceph-ansible when the user is using centos or debian.

If we want to merge, there's no choice but to add tasks to make ceph-ansible work on Yocto. There's even a task that limits the distribution to “Debian, RedHat, ClearLinux, Suse”.

However, I agree that the behavior of ceph-ansible should not be modified to add Yocto support. I will modify the patches accordingly. In particular, the “zap and destroy osds created by ceph-volume with devices” tasks should not be removed, I'll just make them specific to the initial distributions supported by ceph-ansible.

Regarding how patches are applied, this is currently done by the prepare.sh script, which makes no distinction between Yocto/Debian. I think that if we don't change the behavior of ceph-ansible for Debian, that's fine. But the other option is to add a playbook that will apply patches if at least one of the nodes is running Debian. 

What's your opinion on this ?

src/ceph-ansible-patches/distro_not_supported.diff
```diff
@@ -0,0 +1,15 @@
...
+--- a/roles/ceph-validate/tasks/check_system.yml
++++ b/roles/ceph-validate/tasks/check_system.yml
+@@ -22,7 +22,9 @@
+ - name: fail on unsupported distribution
+   fail:
+     msg: "Distribution not supported {{ ansible_facts['os_family'] }}"
+-  when: ansible_facts['os_family'] not in ['Debian', 'RedHat', 'ClearLinux', 'Suse']
++  when:
++    - ansible_facts['os_family'] not in ['Debian', 'RedHat', 'ClearLinux', 'Suse']
++    - "'Yocto' not in ansible_distribution"
```
eroussy, line 12: Do we stick to the ansible_distribution test, or do we have a better system ?
@Revalioli

roles/configure_admin_user/README.md
```diff
@@ -0,0 +1,22 @@
+# Add LiveMigration User Role
```
eroussy, line 1: wrong title

eroussy (changes requested):

eroussy:
One commit message is wrong : 
"admin_user: move the user configuration to a dedicated playbook"
should be changed to "admin_user: move the user configuration to a dedicated *role*"

src/ceph-ansible-patches/distro_not_supported.diff
dupremathieu: We can keep as it and change it inside the PR #645.

dupremathieu (approved):

PR #644, opened 2024-11-29
livemigration_user: use default path for authorized_keys

dupremathieu:
The authorized_keys file is now created in the default path for the livemigration_user. This is done to avoid the need to specify the path in the playbook.


TanguyRaufflet (commented):
I think you can also remove the user creation since it's already done in meta-seapath.

dupremathieu:
> I think you can also remove the user creation since it's already done in meta-seapath.

Yes, we can, but we can also keep it will work as a test.

Revalioli (approved):

PR #645, opened 2024-11-29
Refactor distro detection

Revalioli:
The ansible fact `ansible_distribution` work fine to detect if a Seapath distribution is Debian or CentOS, but doesn't work well for Yocto Seapath.
With Yocto, `ansible_distribution` will be set to the value of DISTRO_NAME. As it is usual to have multiple distro configurations in a Yocto project, with different DISTRO_NAME values, `ansible_distribution` wouldn't have a deterministic value for every Yocto Seapath build.

With yocto scarthgap, the `/etc/os-release` file has a field CPE_NAME which will always contains "openembedded".
It can then be used as a reliable way to detect a Yocto Seapath distribution.

This PR propose a new role, detect_seapath_distro, that will set a fact `seapath_distro` with a value of either Yocto, Debian or CentOS. To detect the correct distro, is uses `ansible_distribution*` facts and the /etc/os-release file.
The new fact is then used everywhere a task needs to know the seapath distribution of the host.


Revalioli:
Note: currently only the last two commits are for this PR. Other commits belongs to https://github.com/seapath/ansible/pull/643, and will be removed when https://github.com/seapath/ansible/pull/643 is merged.

playbooks/seapath_setup_main.yaml
```diff
@@ -9,23 +9,21 @@
...
-  tasks:
-    - name: Gather only the ansible_distribution fact
-      ansible.builtin.setup:
-        filter: ansible_distribution
+  roles:
+    - detect_seapath_distro
 
 - import_playbook: seapath_setup_prerequiscentos.yaml
-  when: ansible_distribution | regex_search("CentOS|RedHat")
+  when: seapath_distro == "CentOS"
```
insatomcat, line 20: import_playbook does not take "when" into account

roles/detect_seapath_distro/tasks/main.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+- name: Detect Centos distribution
+  set_fact:
+    seapath_distro: CentOS
+  when: ansible_distribution | regex_search("CentOS|RedHat")
+
+- name: Detect Yocto distribution
+  block:
+    - name: Search CPE_NAME field of {{ ansible_distribution_file_path }}
+      command: grep 'CPE_NAME.*cpe:/o:openembedded' {{ ansible_distribution_file_path }}
```
dupremathieu, line 22: I am not sure we can rely on CPE_NAME because we don't know how CPEs are defined.

Revalioli: There is an official CPE_NAME dictionary that is used to define CVE if I understood well, but nothing says explicitly that non-official CPE shouldn't be used.
Also `cpe:/o:openembedded` isn't even in this dictionary.

This pattern for the CPE_NAME variable is hardcoded this way in poky, so it probably won't change for scarthgap.
The goal here is only to use what is hardcoded in poky, and not to set the CPE itself, maybe we can use it for now, and if in the future its CPE_NAME changes in poky or Seapath needs its own CPE, we would then adapt the role ?

playbooks/seapath_setup_main.yaml
insatomcat: yes, sorry, I wasn't clear but that's why I meant

Revalioli: What would you propose instead ?
`import_playbook` cannot be used inside a play, so not possible to surround it with a block or another statement on which to put the `when`.
The other way would be to put multiple `when` directly in the playbooks be we would then still get multiple "skipping" messages, and it would make a lot of repetitive `when` conditionals

insatomcat: I would probably go with having separate "main" playbooks for each flavor, or leave it to the user to create its own playbook.
But this is future work and is not related to this PR.

Revalioli:
PR is now rebased on the latest debiancentos commits.

Note that the change from `ansible_distribution` to `seapath_distro` variable was done everywhere in the repo except for specific Debian playbooks and templates (examples: `debian_physical_machine` role, `seapath_setup_prerequisdebian.yaml` playbook, groups and users configuration template files). Only playbooks and roles that are explicitly made for Debian or CentOS are concerned.

If you think `ansible_distribution` should be changed anyway don't hesitate to comment

roles/add_livemigration_user/README.md
```diff
@@ -4,7 +4,7 @@ This role sets the live migration user on the cluster, with ssh key exchanges
 
 ## Requirements
 
-no requirement.
```
eroussy, line 7: Why do you need this role as a dependency ?
There is nothing in the role that uses distro specific variables

insatomcat:
Can you create the meta/main.yml file in this new role so that it's uploadable to galaxy? thx.

roles/ceph_prepare_installation/README.md
```diff
@@ -4,7 +4,7 @@ This role prepares the ceph installation by identifying if it's a new install an
 
 ## Requirements
 
-no requirement.
```
eroussy, line 7: same, why do you need this dependency ?

eroussy (changes requested):

Revalioli:
> Can you create the meta/main.yml file in this new role so that it's uploadable to galaxy? thx.

@insatomcat Thanks, done
Note that I kept the `platforms` section to match other roles, but it seems that recent version of Ansible Galaxy doesn't use this metadata anymore, see https://github.com/ansible/galaxy/issues/3292

roles/add_livemigration_user/README.md
Revalioli: Removed, this was forgot after rebasing on the new debiantcentos branch after merge of https://github.com/seapath/ansible/pull/643

roles/ceph_prepare_installation/README.md
Revalioli: Same as for `roles/add_livemigration_user`

ebail:
Hi @insatomcat ! Have you got some comments on that ? Thanks

eroussy (dismissed):

eroussy (dismissed):

eroussy (approved):

PR #660, opened 2024-12-10
roles/network_resolved: do not disable systemd-resolved

dupremathieu:
The systemd-resolved service is required by libvirt. The service will works even if no DNS server is defined.


insatomcat:
can you please detail the problem? or do you have a link to the bug?

dupremathieu:
> can you please detail the problem? or do you have a link to the bug?

Libvirt can manage VMs network and use dnsmasq to setup a DHCP server for VMs.

This dnsmasq fails if there is no `/etc/resolv.conf` file.

```
Dec 11 10:16:50 adv-ecu-579 dnsmasq[1136]: directory /etc/resolv.conf for resolv-file is missing, cannot poll
Dec 11 10:16:50 adv-ecu-579 dnsmasq[1136]: FAILED to start up
Dec 11 10:16:50 adv-ecu-579 libvirtd[1038]: internal error: Child process (VIR_BRIDGE_NAME=virbr0 /usr/bin/dnsmasq --conf-file=/var/lib/libvirt/dnsmasq/default.conf --leasefile-ro --dhcp-script=/usr/libexec/lib>
                                            dnsmasq: directory /etc/resolv.conf for resolv-file is missing, cannot poll
```

The `/etc/resolv.conf` is created by systemd-resolved so if it is stopped, there is no `/etc/resolv.conf` file.

Keeping systemd-resolved is the easy way to solve the issue. I guess it should also be possible to modify the dnsmasq configuration generated by libvirt to remove the DNS management part and keep only the DHCP.

insatomcat:
On my setup, `/etc/resolv.conf` is a link to a non-existing `../run/systemd/resolve/stub-resolv.conf`.
And all I get in libvirtd log is a simple `Unable to find 'dnsmasq' binary in $PATH: No such file or directory`.
I don't think we need to keep systemd-resolved running all the time, we just need to create any empty /etc/resolv.conf file at setup.

dupremathieu:
> On my setup, resolved is disabled and stopped, and `/etc/resolv.conf` is a link to a non-existing `../run/systemd/resolve/stub-resolv.conf`. And all I get in libvirtd log is a simple, benign`Unable to find 'dnsmasq' binary in $PATH: No such file or directory`. I don't think we need to keep systemd-resolved running all the time, we just need to create any empty /etc/resolv.conf file at setup. Or at least condition resolved to the need of having dnsmasq (which I'm pretty sure won't be a production need...)

Create an empty /etc/resolv.conf should fix the issue as well.

Regarding `dnsmasq` it is used for internal network with IPs managed by libvirt.
We don't have this use case in our case, but the libvirt creates some default network which use `dnsmasq` (at least on Yocto).

So, we have two options:
* support the usecase: install `dnsmasq` and create an empty /etc/resolv.conf.
* not support the usecase: remove `dnsmasq` and the default network.

To me, I think we can drop this usecase. What do you think @insatomcat?

insatomcat:
I can't recall if it's something I did something for or if it's by default, but on debian the "default" libvirt network is inactive at deployment (I checked on the CI):
```
admin@ci1deb12:~$ virsh net-list --all
 Name      State      Autostart   Persistent
----------------------------------------------
 default   inactive   no          yes
```
Do not having dnsmasq is not a problem, and we do not install the package.
If you ask me it works well as is, so maybe you can make yocto not having dnsmasq as well.

PR #663, opened 2024-12-11
Improve latency tests

Revalioli:
- Run latency test for 5H
- Save only the latency of the first SV of each 4000 iteration as a temporary workaround due to current memory limitations


dupremathieu:
In this case, you can just create the branch to run the CI in the action tab. No need to create the PR.

PR #666, opened 2024-12-13
debian_hypervisor: synchronize team0_x port state with OVS

ycongal-smile:
On hypervisor shutdown, OVS is stopped a long time before the team0_x
ports are shutdown as well.

In the meantime:
* The switching done by OVS is stopped
* The corosync totem stops circulating and nodes are disconnected from
  the pacemake cluster which leads to resources being stopped.
* After a while, RTSP detects this and change ports configuration to
  avoid the shutdown node and corosync/pacemaker start back with the
  left nodes.

To avoid this, use systemd services to shutdown team0_x ports as soon as
OVS is stopped to signal RTSP to change the configuration "instantly"
and thus, avoid the corosync totem timeout.

This PR also include preliminary work and a unrelated readme fix

TODO:

- [x] Test feature as deployed
- [x] Test ansible deployment


insatomcat (approved):

Paullgk:
Hello, 
This PR is introducing a regression when configuring a Debian standalone distribution, as `team0` network is not used.

PR #667, opened 2024-12-16
roles/add_kernel_parameters: create the role

dupremathieu:
This role will add kernel parameters to the existing ones. This was previously done in the ci_restore_snapshot role.


roles/add_kernel_parameters/tasks/main.yml
```diff
@@ -0,0 +1,50 @@
+# Copyright (C) 2024 RTE
+# Copyright (C) 2024 Savoir-faire Linux, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+---
+
```
eroussy, line 6: You should add a check to know if the kernel parameters is already present. 
Like it is done here : https://github.com/seapath/ansible/blob/debiancentos/roles/yocto/kernel_params/tasks/main.yaml#L9

roles/add_kernel_parameters/tasks/main.yml
```diff
@@ -0,0 +1,50 @@
...
+  with_items:
+    - ipv6.disable=1
+    - efi=runtime
+    - "{{ grub_append | default([]) }}"
+- name: Update grub
+  command: "{{ grub_update_command }}"
+
+- name: Umount the boot partition if needed
+  command: /usr/share/update/mount_boot.sh umount
+  when: mount_boot_script.stat.exists
```
eroussy, line 50: Is this role in conflict with `yocto/kernel_params` ? 
Should  we remove the kernel_params role ?

roles/add_kernel_parameters/tasks/main.yml
```diff
@@ -0,0 +1,50 @@
...
+    path: /usr/sbin/update-grub
+  register: grub_update_command
+  when: grub_update_command is not defined
+
+- name: Set the grub update command to "update-grub"
+  set_fact:
+    grub_update_command: "update-grub"
+  when:
+    - grub_update_command is not defined
+    - not grub_update_command.stat.exists
```
eroussy, line 27: Why ?
if grub_update_command.stat.exists is false, it means that "update-grub" binary doesn't exists

eroussy (changes requested):

PR #678, opened 2025-01-06
Bug fixs from CI testing

yairpod:
1. When restoring the LVM snapshot the snapshot partition is not immediately gone after reboot, a race condition exists between the ansible runner getting to the "Recreate the snapshot" stage and the target machine syncing the LVM partitions and removing the snapshot, on a slower machine this will cause the ansible to try and create an root-snapshot partition and receiving the error  "Operation not permitted on hidden LV vg1/root-snap", the solution is to wait for the root-snap partition to be cleared.
2. CentOS natively uses NetworkManager as the network service, SEAPATH does not at the moment support this, we use systemd-networkd, for the network to be properly configured NetworkManager must be shot down and systemd-networkd enabled and started, if that was not the case to begin with, a restart will be needed.

### DIFF
--- a/roles/centos/tasks/main.yml
+++ b/roles/centos/tasks/main.yml
@@ -201,3 +201,23 @@
 - name: update-grub
   command: grub2-mkconfig -o /boot/grub2/grub.cfg
   when: updategrub1.changed or updategrub2.changed
+
+- name: Stop and Disable NetworkManager
+  service:
+    name: NetworkManager
+    state: stopped
+    enabled: no
+  register: nm_stop
+
+- name: Start and Enable systemd-networkd
+  service:
+    name: systemd-networkd
+    state: started
+    enabled: yes
+  register: sd_start
+
+
+- name: Set need_reboot to true if any change was made
+  set_fact:
+    need_reboot: true
+  when: nm_stop is changed or sd_start is changed
--- a/roles/ci_restore_snapshot/tasks/main.yml
+++ b/roles/ci_restore_snapshot/tasks/main.yml
@@ -25,6 +25,14 @@
 - name: Refresh LVM
   command:
     cmd:  lvchange --refresh vg1
+
+- name: wait until snapshot is cleared
+  command: lvs -a
+  register: result
+  until: result.stdout | regex_search('root-snap', ignorecase=True) is none
+  retries: 6
+  delay: 10
+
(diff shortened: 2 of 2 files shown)


eroussy:
Hi Yair
There are many warnings in ansible-llint that we specifically ignore, because we didn't take the time (for now)
The only actual error that you have is this one, that does impact your files :
```
command-instead-of-shell: Use shell only when shell functionality is required
roles/ci_restore_snapshot/tasks/main.yml:29 Task/Handler: wait until snapshot is cleared
```

yairpod:
> Hi Yair There are many warnings in ansible-llint that we specifically ignore, because we didn't take the time (for now) to correct them. The only actual error that you have is this one, that does impact your files :
> 
> ```
> command-instead-of-shell: Use shell only when shell functionality is required
> roles/ci_restore_snapshot/tasks/main.yml:29 Task/Handler: wait until snapshot is cleared
> ```

This functionality is required, I could not find a non-shell ansible command that would do the task

yairpod:
@eroussy I can not find a non-shell way to achieve the same effect

eroussy:
> @eroussy I can not find a non-shell way to achieve the same effect

I don't think that is the problem. I think you just have to use `command: lvs -a` instead of `ansible.builtin.shell: lvs -a`

roles/centos/tasks/main.yml
```diff
@@ -201,3 +201,23 @@
...
   when: updategrub1.changed or updategrub2.changed
+
+- name: Stop and Disable NetworkManager
+  service:
+    name: NetworkManager
+    state: stopped
+    enabled: no
+  register: nm_stop
+
+- name: Start anad Enable systemd-networkd
```
dupremathieu, line 212: anad -> and

dupremathieu (changes requested):

roles/centos/tasks/main.yml
yairpod: Amended

eroussy (approved):

dupremathieu (approved):

eroussy:
Hi Yair
As mentioned on the slack, we are currently working on the Yocto CI, so we had to disable it.
You will not be able to merge your PR now, but don't worry, I will rebase it and merge it myself tomorrow.

PR #679, opened 2025-01-08
cluster_setup_ceph: disable PrivateTmp for ssh during Ceph configuration

Revalioli:
Commit [1] in meta-seapath enabled the PrivateTmp systemd option for every ssh sessions with SEAPATH Yocto.
This means that every ssh sessions, including ansible ones, will use a separate mount namespace. Any new files in /tmp or new tmpfs during ansible tasks are deleted once the ssh session closes.

This is a problem regarding Ceph and ceph-ansible. Directories in /var/lib/ceph/osd/, which contains multiple OSD files, are in fact tmpfs created by ceph-ansible. With the PrivateTmp option, these directories are then lost once the Ansible ssh session closes.

As a workaround, temporarily disable the PrivateTmp option for the Ceph configuration.
Note that this only affects SEAPATH Yocto.

[1]: https://github.com/seapath/meta-seapath/commit/22cddd356d2dbb387c76049ae80eb91a982f8082

### DIFF
--- a/playbooks/cluster_setup_ceph.yaml
+++ b/playbooks/cluster_setup_ceph.yaml
@@ -29,7 +29,45 @@
   roles:
     - ceph_expansion_lv
 
+- name: Ceph OSD hosts pre-configuration
+  hosts: osds
+  tasks:
+    - name: Disable private tmpfs for ceph configuration
+      when: seapath_distro == "Yocto"
+      block:
+        - name: Create /run/systemd/system/sshd@.service.d directory
+          file:
+            path: /run/systemd/system/sshd@.service.d
+            state: directory
+        - name: Temporarily disable private tmpfs for ssh sessions
+          copy:
+            dest: /run/systemd/system/sshd@.service.d/99-temp-disable-private-tmp.conf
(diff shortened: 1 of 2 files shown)


eroussy (dismissed):

playbooks/cluster_setup_ceph.yaml
```diff
@@ -29,8 +29,30 @@
   roles:
     - ceph_expansion_lv
 
+- name: Ceph OSD hosts pre-configuration
+  hosts: osds
+  tasks:
+    - name: Temporarily disable private tmpfs for ssh sessions
+      lineinfile:
+        path: /usr/lib/systemd/system/sshd@.service.d/sshd-private-tmp.conf
```
dupremathieu, line 37: This file id in ro on Yocto.

Use an systemd override file instead.

dupremathieu (approved):

dupremathieu (changes requested):
See the comment.

Revalioli:
Problem fixed in https://github.com/seapath/meta-seapath/pull/302

PR #685, opened 2025-01-13
ci_yocto: add support for Aaeon I.MX8 board

TanguyRaufflet:
The reboot_on_usb_drive have been replaced by clean_arm_machines on arm machines. This is mandatory since there is no solution to flash the board yet.

The cyclictest have been removed since the kernel is not RT.

Get_system_info have been modified so that multiple system_info can be generated with different names.

### DIFF
--- a/playbooks/ci_all_machines_tests.yaml
+++ b/playbooks/ci_all_machines_tests.yaml
@@ -15,6 +15,7 @@
 - hosts:
     - cluster_machines
     - standalone_machine
+  gather_facts: false
   roles:
     - ci_yocto/get_system_info
 ...
--- a/playbooks/ci_all_machines_tests.yaml
+++ b/playbooks/ci_all_machines_tests.yaml
@@ -8,6 +8,7 @@
     - cluster_machines
     - standalone_machine
     - VMs
+  gather_facts: false
   roles:
     - ci_yocto/run_tests
 
--- a/playbooks/ci_standalone_setup.yaml
+++ b/playbooks/ci_standalone_setup.yaml
@@ -1,4 +1,4 @@
-# Copyright (C) 2023-2024 Savoir-faire Linux, Inc.
+# Copyright (C) 2023-2025 Savoir-faire Linux, Inc.
 # SPDX-License-Identifier: Apache-2.0
 
 ---
@@ -11,5 +11,14 @@
   roles:
      - role: ci_yocto/reboot_on_usb_drive
        when: ansible_facts['architecture'] == "x86_64"
+
+- name: Swupdate and wipe persistent data
+  hosts:
+    - standalone_machine
+  gather_facts: true
+  become: true
+  roles:
+      - role: ci_yocto/clean_arm_machines
+        when: ansible_facts['architecture'] == "aarch64"
 
 - import_playbook: seapath_setup_main.yaml
--- a/playbooks/ci_standalone_setup.yaml
+++ b/playbooks/ci_standalone_setup.yaml
@@ -7,8 +7,9 @@
   hosts:
     - cluster_machines
     - standalone_machine
-  gather_facts: false
+  gather_facts: true
   roles:
-    - ci_yocto/reboot_on_usb_drive
+     - role: ci_yocto/reboot_on_usb_drive
+       when: ansible_facts['architecture'] == "x86_64"
 
 - import_playbook: seapath_setup_main.yaml
--- a/roles/ci_yocto/clean_arm_machines/tasks/main.yaml
+++ b/roles/ci_yocto/clean_arm_machines/tasks/main.yaml
@@ -1,0 +1,40 @@
+# Copyright (C) 2025 Savoir-faire Linux, Inc.
+# SPDX-License-Identifier: Apache-2.0
+#
+# This role allows to clean the hypervisor.
+# It performs a software update and removes persistent data.
+
+---
+- name: Create /tmp/update folder
+  file:
+    path: /tmp/update
+    state: directory
+    mode: '0755'
+- name: Mount the USB key
+  mount:
+    src: "{{ usb_device }}"
+    path: /tmp/update
+    fstype: vfat
+    state: mounted
+- name: Execute Swupdate
+  command: /usr/bin/swupdate -i /tmp/update/{{ swupdate_image }}
+- name: Unmount the USB key
+  mount:
+    path: /tmp/update
+    state: unmounted
+- name: Get overlay mounts
+  command: awk '$3 == "overlay" {print $2}' /proc/mounts
+  register: overlay_mounts
+- name: Unmount overlays
+  command: umount -l {{ item }}
+  with_items: "{{ overlay_mounts.stdout_lines }}"
+- name: Remove persistent data
+  file:
+    path: '{{ item }}'
+    state: absent
+  with_items:
+    - /mnt/persistent/etc
+    - /mnt/persistent/home
+    - /mnt/persistent/var
+- name: Reboot the system
+  reboot:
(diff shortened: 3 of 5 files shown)


roles/ci_yocto/run_tests/tasks/main.yaml
```diff
@@ -23,11 +23,14 @@
...
+- name: Cyclictest
+  block:
+    - name: Execute cyclitect script
+      script: run_cyclictest.py '{{ tmp_test_dir.path }}/cyclictest.txt'
+    - name: Fetch results
+      fetch:
+        src: '{{ tmp_test_dir.path }}/cyclictest.txt'
+        dest: "{{ cukinia_test_prefix | default('..') }}/cyclictest_{{ inventory_hostname }}.txt"
+        flat: yes
+  when: ansible_facts['hostname'] != "yoctoCI-aaeon"
```
eroussy, line 35: I think it would be better to add a variable in the inventory
I propose
```
when: skip_cyclictest is not defined or not skip_cyclictest
```
Then you just have to add this line in yocto-ci-aaeon inventory : 
```
skip_cyclictest: true
```

roles/ci_yocto/reboot_on_usb_drive/tasks/main.yaml
```diff
@@ -5,11 +5,14 @@
...
+      script: "configure_boot_next_usb_drive.sh"
+      register: result
+      changed_when: result.rc == 0
+      failed_when: result.rc == 1 or result.rc > 2
 
-- name: Reboot the system
-  reboot:
+    - name: Reboot the system
+      reboot:
+  when: ansible_facts['architecture'] == "x86_64"
```
eroussy, line 18: Add this condition directly in `playbooks/ci_standalone_setup.yaml` when the role is called : 
```
- name: Configure next reboot to boot on USB drive
  hosts:
    - cluster_machines
    - standalone_machine
  gather_facts: false
  gather_facts: true
  roles:
     - role: ci_yocto/reboot_on_usb_drive
       when: ansible_facts['architecture'] == "x86_64"

File: roles/ci_yocto/get_system_info/tasks/main.yaml [hunk 3]
```diff
@@ -13,6 +13,6 @@
 - name: Fetch results
   fetch:
     src: '{{ tmp_test_dir.path }}/system_info.adoc'
-    dest: "{{ cukinia_test_prefix | default('..') }}/system_info.adoc"
+    dest: "{{ cukinia_test_prefix | default('..') }}/system_info_{{ ansible_facts['hostname'] }}.adoc"
```
eroussy, line 16: Just for you to know, a simplest way to do that :  `{{ inventory_hostname }}`
It take the name of the board described in the inventory instead of the hostname of the board (it should be the same in our case)
https://docs.ansible.com/ansible/latest/reference_appendices/special_variables.html#term-inventory_hostname

It is not really important, so you can let that as it is, if you want.

roles/ci_yocto/clean_arm_machines/tasks/main.yaml
```diff
@@ -0,0 +1,43 @@
...
+      file:
+        path: '{{ item }}'
+        state: absent
+      with_items:
+        - /mnt/persistent/etc
+        - /mnt/persistent/home
+        - /mnt/persistent/var
+    - name: Reboot the system
+      reboot:
+  when: ansible_facts['architecture'] == "aarch64"
```
eroussy, line 43: Same remark, I prefer to put this condition in the playbook `ci_standalone_setup`

roles/ci_yocto/clean_arm_machines/tasks/main.yaml
```diff
@@ -0,0 +1,43 @@
...
+        state: directory
+        mode: '0755'
+    - name: Mount the USB key
+      mount:
+        src: /dev/sda1
+        path: /tmp/update
+        fstype: vfat
+        state: mounted
+    - name: Execute Swupdate
+      command: /usr/bin/swupdate -i /tmp/update/seapath-host-docker-imx8p-swu-image-seapath-hypervisor-aaeon-2g.swu
```
eroussy, line 22: I would prefer to put that in a variable in the inventory

roles/ci_yocto/clean_arm_machines/tasks/main.yaml
```diff
@@ -0,0 +1,43 @@
...
+- name: Clean arm hypervisor
+  block:
+    - name: Create /tmp/update folder
+      file:
+        path: /tmp/update
+        state: directory
+        mode: '0755'
+    - name: Mount the USB key
+      mount:
+        src: /dev/sda1
```
eroussy, line 17: I would prefer to put that in a variable in the inventory

eroussy (changes requested):

roles/ci_yocto/reboot_on_usb_drive/tasks/main.yaml
TanguyRaufflet: done

roles/ci_yocto/get_system_info/tasks/main.yaml
TanguyRaufflet: Done

roles/ci_yocto/clean_arm_machines/tasks/main.yaml
TanguyRaufflet: Done

roles/ci_yocto/clean_arm_machines/tasks/main.yaml
TanguyRaufflet: Done

roles/ci_yocto/clean_arm_machines/tasks/main.yaml
TanguyRaufflet: Done

playbooks/ci_all_machines_tests.yaml
```diff
@@ -8,6 +8,7 @@
     - cluster_machines
     - standalone_machine
     - VMs
+  gather_facts: true
```
eroussy, line 11: You don't need the facts anymore

roles/ci_yocto/reboot_on_usb_drive/tasks/main.yaml
```diff
@@ -13,3 +13,4 @@
 
 - name: Reboot the system
   reboot:
+
```
eroussy, line 16: Remove this line

playbooks/ci_standalone_setup.yaml
```diff
@@ -9,6 +9,7 @@
     - standalone_machine
   gather_facts: false
```
eroussy, line 10: I think you need the facts here

playbooks/ci_all_machines_tests.yaml
```diff
@@ -15,6 +15,7 @@
 - hosts:
     - cluster_machines
     - standalone_machine
+  gather_facts: true
```
eroussy, line 18: You don't need the facts here

eroussy (changes requested):

playbooks/ci_all_machines_tests.yaml
TanguyRaufflet: Done

playbooks/ci_standalone_setup.yaml
TanguyRaufflet: Done

playbooks/ci_all_machines_tests.yaml
TanguyRaufflet: Done

roles/ci_yocto/reboot_on_usb_drive/tasks/main.yaml
TanguyRaufflet: Done

eroussy (dismissed):

eroussy (approved):

PR #690, opened 2025-01-17
debian_hypervisor: add missing team0 condition

Paullgk:
Currently, when deploying a SEAPATH Debian standalone configuration some tasks failed has it requires team0 network, which is not used on a standalone configuration (only cluster).
This issue has been introduced with commit
3691fb2f8896c573a962b8ee3714340e53e7d2d7 from PR
https://github.com/seapath/ansible/pull/666.

A missing `when` condition is needed to run these tasks only if team0 network is used.

### DIFF
--- a/roles/debian_hypervisor/tasks/main.yml
+++ b/roles/debian_hypervisor/tasks/main.yml
@@ -210,6 +210,7 @@
     group: root
     mode: 0644
   register: ovsvswitchd
+  when: team0_0 is defined and team0_1 is defined
 - name: Copy team0_x@.service file
   ansible.builtin.copy:
     src: team0_x@.service
@@ -218,6 +219,7 @@
     group: root
     mode: 0644
   register: team0_x_service
+  when: team0_0 is defined and team0_1 is defined
 - name: Restart ovs-vswitchd
   ansible.builtin.systemd:
     state: restarted


eroussy (dismissed):

Revalioli (dismissed):

roles/debian_hypervisor/tasks/main.yml
```diff
@@ -210,6 +210,7 @@
     group: root
     mode: 0644
   register: ovsvswitchd
+  when: team0_0 is defined
```
ycongal-smile, line 213: Is it possible to use `host: cluster_machines` ([ansible pattern](https://docs.ansible.com/ansible/latest/inventory_guide/intro_patterns.html)) ?

I'm afraid that testing team0_0 definition might be too fragile.

Paullgk: I don't think it might be better as the task is called within a role, for host `standalone_machine` and `hypervisors`, and so must be compatible even if hypervisors host is in `cluster_machines` or `standalone_machine` groups.
`team0_0` is a documented var is the example inventory and SEAPATH documentation, and used when we setup a cluster.
Otherwise another solution would be to move the tasks added with https://github.com/seapath/ansible/pull/666 in a cluster role, such as `roles/network_clusternetwork/tasks/main.yml`

ycongal-smile: >  team0_0 is a documented var is the example inventory and SEAPATH documentation, and used when we setup a cluster.

Yes, you are right. Ok, I'm fine with merging with the team0_0 test, but should it be
```suggestion
  when: team0_0 is defined and team0_1 is defined
```
? (since both variables are used in template)

roles/debian_hypervisor/tasks/main.yml
```diff
@@ -210,6 +210,7 @@
     group: root
     mode: 0644
   register: ovsvswitchd
+  when: team0_0is defined and team0_1 is defined
```
ycongal-smile, line 213:
```suggestion
  when: team0_0 is defined and team0_1 is defined
```

dupremathieu (approved):

PR #692, opened 2025-01-17
ci_latency_tests: add latency tests for ARM CI

TanguyRaufflet:
This commit allows starting sv_timestamp_logger on the publisher, the VMs and the I.MX8P board using the right architecture.

This PR also contains the modifications provided by https://github.com/seapath/ansible/pull/694 as they cannot be merged separately.


playbooks/ci_latency_tests.yaml
```diff
@@ -14,7 +15,7 @@
           mode: '0777'
       - name: "Copy archive"
         copy:
-          src: "../ci_latency_tests/build/sv_timestamp_logger.tar"
+          src: "{{ '../ci_latency_tests/build/sv_timestamp_logger_' + ('amd64' if inventory_hostname != 'yoctoCI-aaeon' else 'arm64') + '.tar' }}"
```
eroussy, line 18: I would prefer to use a inventory variable : 
`"../ci_latency_tests/build/{{ sv_timestamp_logger_image }}"`

playbooks/ci_latency_tests.yaml
```diff
@@ -58,7 +59,7 @@
   become: true
   tasks:
     - name: Load Docker archive
-      command: "docker image load -i /tmp/latency_tests/sv_timestamp_logger.tar"
+      command: "docker image load -i /tmp/latency_tests/sv_timestamp_logger_amd64.tar"
```
eroussy, line 62: Don't modify that
see previous command

playbooks/ci_latency_tests.yaml
```diff
@@ -14,7 +15,7 @@
           mode: '0777'
       - name: "Copy archive"
         copy:
-          src: "../ci_latency_tests/build/sv_timestamp_logger.tar"
+          src: "{{ '../ci_latency_tests/build/sv_timestamp_logger_' + ('amd64' if inventory_hostname != 'yoctoCI-aaeon' else 'arm64') + '.tar' }}"
           dest: "/tmp/latency_tests/"
```
eroussy, line 19: It is better to keep the name "sv_timestamp_logger.tar" on the machines, no matter the architecture.
We know the architecture, because it is the one of the publisher (amd64)

You have to use the right file (arm64 or amd64) on the runner, but please name it sv_timestamp_logger.tar on the publisher
`dest: "/tmp/latency_tests/sv_timestamp_logger.tar"`

playbooks/ci_latency_tests.yaml
```diff
@@ -70,7 +71,7 @@
         -v /tmp:/tmp
         --cap-add=NET_ADMIN
         --cap-add=SYS_NICE
-        sv_timestamp_logger
+        sv_timestamp_logger_amd64
```
eroussy, line 74: Same remark. 
Lets keep the docker image name "sv_timestamp_logger" on each machine.
We don't need to know the architecture of the machine. We only need to send the right archive to each machine.

playbooks/ci_latency_tests.yaml
```diff
@@ -92,7 +93,27 @@
         --network host
         -v /tmp:/tmp
         --cap-add=SYS_NICE
-        sv_timestamp_logger
+        sv_timestamp_logger_amd64
+        -d {{ sv_interface }}
+        -f "/tmp/latency_tests/ts_{{ inventory_hostname }}.txt"
+        -s svID0000
+
+- name: Launch sv timestamp logger on arm hypervisors
```
eroussy, line 101: This section is not useful :
If your dockers are named "sv_timestamp_logger" on each machines, you can just add "yoctoCI-aaeon" host on the previous block of tasks.
You will then use the same code for the VMs and aaeon board.

eroussy (changes requested):

playbooks/ci_latency_tests.yaml
TanguyRaufflet: Done

playbooks/ci_latency_tests.yaml
TanguyRaufflet: Done

playbooks/ci_latency_tests.yaml
TanguyRaufflet: Done

playbooks/ci_latency_tests.yaml
TanguyRaufflet: Done

eroussy (approved):

playbooks/ci_latency_tests.yaml
TanguyRaufflet: Done

dupremathieu (changes requested):
I prefer not replacing the Aaeon name ny ARM CI.

TanguyRaufflet:
> I prefer not replacing the Aaeon name ny ARM CI.

But if we want to add a second Arm machine to the CI, we'll have to give it a different name. That's why I thought using Aaeon instead of Arm was a better choice.

eroussy (approved):

dupremathieu (changes requested):
ci_latency_tests: add latency tests for Arm -> ci_latency_tests: add latency tests for ARM CI

eroussy (approved):

dupremathieu (approved):

PR #694, opened 2025-01-17
seapath_setup_timemaster: remove ptp_status_vsock for ARM CI

TanguyRaufflet:
As mentioned in the vsock documentation, this address family is used to ease communication between VMs and host. Since the I.MX8 board used on the CI is using a SEAPATH host container image, this service cannot be enabled.


eroussy (approved):

insatomcat:
Can I ask what you mean by "containers" and the relation to vsock?

playbooks/seapath_setup_timemaster.yaml
```diff
@@ -8,4 +8,13 @@
...
+
+- name: Configure ptp_vsock
+  hosts:
+    - cluster_machines
+    - standalone_machine
+  become: true
+  gather_facts: true
+  roles:
+    - role: ptp_status_vsock
+      when: ansible_facts['architecture'] == "x86_64"
```
dupremathieu, line 20: The CPU architecture has nothing to do with whether you choose to have vsock or not.
Use a container only variable or something else.

dupremathieu (changes requested):
Replace Aaeon by container only.

TanguyRaufflet:
> Can I ask what you mean by "containers" and the relation to vsock?

Host container images do not support virtualization, they will simply run Docker containers instead of VMs. So there's no reason to use vsock, since its purpose is to facilitate VM/hypervisor communication.

playbooks/seapath_setup_timemaster.yaml
TanguyRaufflet: Done.

playbooks/seapath_setup_timemaster.yaml
```diff
@@ -8,4 +8,13 @@
...
+
+- name: Configure ptp_vsock
+  hosts:
+    - cluster_machines
+    - standalone_machine
+  become: true
+  gather_facts: true
+  roles:
+    - role: ptp_status_vsock
+      when: container_only is not defined or not container_only
```
eroussy, line 20: If you define a new variable, please document it `roles/ptp_status_vsock/README.md`

eroussy (changes requested):

playbooks/seapath_setup_timemaster.yaml
TanguyRaufflet: Done

roles/ptp_status_vsock/README.md
```diff
@@ -8,7 +8,9 @@ No requirement.
 
 ## Role Variables
 
-No variable.
+| Variable        | Requiered | Type           | Comments                                                           |
```
eroussy, line 11: Required

eroussy (changes requested):

eroussy (approved):

playbooks/seapath_setup_timemaster.yaml
dupremathieu: This is not a role variable. We should provide a container only inventory example.

Before creating this variable, we should think how we want to integrate the no virtualization case.
There are several solutions:
- move virtualization stuff into dedicated playbooks. I think this is already the case, except for this playbook
- create a dedicated in host group
- use a variable

Personally, I prefer the playbook split solution.

What do you think @eroussy @insatomcat?

dupremathieu (changes requested):
Replace ARM by container only or no virtualization.

roles/ptp_status_vsock/README.md
```diff
@@ -8,7 +8,9 @@ No requirement.
...
+|-----------------|-----------|----------------|--------------------------------------------------------------------|
+| container_only  | no        | Bool           | Set this variable to true to disable ptp_status_vsock installation |
```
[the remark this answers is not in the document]
TanguyRaufflet: Done

TanguyRaufflet:
This PR has been combined with https://github.com/seapath/ansible/pull/692.

PR #696, opened 2025-01-18
Debian: fix renaming of default user from virtu to admin

insatomcat:
following https://github.com/seapath/build_debian_iso/pull/120


insatomcat:
It's a feature to be able to change the admin username without having to reinstall all your machines.
We have a lot of servers and when the project decided we must use "mco" instead of "virtu", it's ansible's job to make it happen. If tomorrow the project decides to go back to admin, or to "rspace" or whatever, we can easily deal with that.

It's also useful when you install debian by something else than build_debian_iso (for example on a raspberry pi where the default user is "pi"), we need ansible to deal with that. I always say that build_debian_iso is a helper, but not required in the workflow. You can install debian any way you like (a month ago you had to, if you wanted arm64 for example).

As ansible can't easily rename an account, we need to remove it (only if the old username is != the desired username) and recreate it.

eroussy:
Ack.
It will be good to add that in SEAPATH features list.
May be on the release note or on the wiki.

eroussy (approved):

PR #697, opened 2025-01-20
vm_manager packaging update

ycongal-smile:


insatomcat:
thanks, this will be taken over by #699

PR #698, opened 2025-01-20
fix bug in loops on cluster hypervisors

insatomcat:
Following 2ffef3b1b0da314798efe6bb5802368ae7f69030

When running tasks on cluster hypervisors (members of cluster_machines and hypervisors groups), any "looped" action must also be restricted to the same intersection of groups

Also we put the when condition before the block keyword, which is an ansible standard enforced by newest version of ansible-lint.


eroussy (commented):
Content is great, but your message commit is too short.
You can put the description of the PR directly in the commit message.

insatomcat:
> Content is great, but your message commit is too short. You can put the description of the PR directly in the commit message.

done

eroussy (approved):

dupremathieu (approved):

PR #701, opened 2025-01-22
Ansible Lint: adapt to newest version

insatomcat:
Moving to newest version of ceph and then ansible will make seapath use the newest version of ansible lint, for which the current configuration reports more than a thousands errors.
We choose to add the "fqcn-builtins" exception (and spare 500 errors that way) and then adapt the code to solve the other 500-600 errors.

### DIFF
--- a/playbooks/ci_latency_tests.yaml
+++ b/playbooks/ci_latency_tests.yaml
@@ -9,46 +9,46 @@
       - yoctoCI-aaeon
   tasks:
       - name: "Create latency_tests directory if it doesn't exist"
-        file:
+        ansible.builtin.file:
           path: "/tmp/latency_tests/"
           state: directory
           mode: '0777'
       - name: "Copy archive"
-        copy:
+        ansible.builtin.copy:
           src: "../ci_latency_tests/build/{{ sv_timestamp_logger_image }}"
           dest: "/tmp/latency_tests/sv_timestamp_logger.tar"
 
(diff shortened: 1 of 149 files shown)


dupremathieu (dismissed):

playbooks/ci_latency_tests.yaml
```diff
@@ -9,46 +9,46 @@
...
   tasks:
       - name: "Create latency_tests directory if it doesn't exist"
-        file:
+        ansible.builtin.file:
           path: "/tmp/latency_tests/"
           state: directory
           mode: '0777'
       - name: "Copy archive"
-        copy:
-          src: "../ci_latency_tests/build/{{ sv_timestamp_logger_image }}"
```
TanguyRaufflet, line 18: This should not be modified, as the file sv_timestamp_logger.tar no longer exists. It has been replaced by sv_timestamp_logger_amd64.tar and sv_timestamp_logger_arm64.tar depending on the target platform.

TanguyRaufflet (changes requested):

playbooks/ci_latency_tests.yaml
insatomcat: this must be a conflict error during rebase, I'll change it back

TanguyRaufflet (dismissed):

dupremathieu (approved):

PR #711, opened 2025-02-04
cqfd: add export flavor

dupremathieu:
The export flavor creates a tarball of the project including the all the file fetch by the prepare.sh script.


eroussy (changes requested):
You should use the archive feature of cqfd IMO

dupremathieu:
> You should use the archive feature of cqfd IMO

It does not work: `tar: .: file changed as we read it`

To avoid that, I create the archive in /tmp and move it after.
It is impossible to do that with the `cqfd release` feature

eroussy (approved):

PR #717, opened 2025-02-09
use deb822 repo format

insatomcat:
This commits changes the way the debian repos are verified, according to the deb822 format.
https://repolib.readthedocs.io/en/latest/deb822-format.html


insatomcat:
arf, the ansible deb822 module is not available before ansible-core 2.15.
This will have to wait for debian 13 seapath release.

PR #724, opened 2025-02-15
adds oraclelinux to seapath supported distros

insatomcat:
This PR makes Seapath able to deploy on OracleLinux Servers. It includes:
- a kickstart file for building the iso
- a cephadm role so that we do not really on distro-specific ceph packages
- ceph 19
- minimal cukinia tests in a specific oraclelinux_tests role
- an operational CI infrastructure, that deploys a cluster + a standalone + 3 VMs and tests everything

It does not include:
- port of the "hypervisor" role, since OracleLinux does not come with an RT kernel. Using the centos RT kernel is an option but not in the scope of this PR.

### DIFF
--- a/.github/workflows/ci-oraclelinux.yml
+++ b/.github/workflows/ci-oraclelinux.yml
@@ -1,0 +1,59 @@
+# Copyright (C) 2023, RTE (http://www.rte-france.com)
+# SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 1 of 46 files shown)


dupremathieu (commented):
LGTM, but please wait for the TSC presentation before merging.

dupremathieu (approved):

dupremathieu (approved):

PR #727, opened 2025-02-15
timemaster: harmonize role variable names

insatomcat:


roles/timemaster/templates/timemaster.conf.j2
```diff
@@ -37,8 +37,8 @@ slaveOnly             1
 
 # IEC 61850-9-3 Profile
 # (from : https://en.wikipedia.org/wiki/IEC/IEEE_61850-9-3)
-network_transport     {{ ptp_network_transport | default ('L2') }}
-delay_mechanism       {{ ptp_delay_mechanism | default ('P2P') }}
+network_transport     {{ timemaster_ptp_network_transport | default ('L2') }}
```
dupremathieu, line 40: We should also move the default value inside the role default directory.

dupremathieu (changes requested):

roles/timemaster/templates/timemaster.conf.j2
insatomcat: done

PR #729, opened 2025-02-17
ansible.cfg: run --skip-tags "package-install" by default

dupremathieu:
Configure ansible.cfg to run each ansible-playbook commands with --skip-tags "package-install" set by default.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -291,13 +291,11 @@ cluster_machines Ansible group.
 
 To launch the playbook _seapath_setup_main.yaml_ use the following command:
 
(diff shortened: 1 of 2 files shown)


eroussy (changes requested):
Please update README accordingly

dupremathieu:
> good idea! but this tag being only use by ceph-ansible, couldn't we set it more "locally" than in the ansible.cfg ? Like in the import_playbook ceph-ansible parts ?

It should be great, but I think it is impossible.

TanguyRaufflet (approved):

PR #738, opened 2025-02-22
ci12 virt

insatomcat:


templates/vm/guest.xml.j2
```diff
@@ -201,7 +201,9 @@
         <console type="pty">
             <target type="serial" port="0" />
         </console>
-        <memballoon model="none" />
+        <memballoon model="virtio">
```
dupremathieu, line 204: Ok but it should not be enable when rt feature in set.
```
{% if "rt" in vm.vm_features %}
<memballoon model="none" />
{% else %}
<memballoon model="virtio">
  <stats period="5" />
</memballoon>
{% endif %}
```

roles/ci_cleanup_syslog/tasks/main.yml
```diff
@@ -2,7 +2,26 @@
...
+    - /var/log/audit
+    - /var/log/openvswitch
+- name: recreate useful folders (2)
   file:
-    state: absent
-    path: /var/log/syslog-ng
+    path: "{{ item }}"
+    state: directory
+    mode: 0755
+    owner: hacluster
```
dupremathieu, line 24: hacluster and haclient are not present in standalone

dupremathieu (changes requested):

dupremathieu (dismissed):

dupremathieu (approved):

PR #740, opened 2025-02-24
debian_physical_machine: Add grub based boot counting and recovery

ycongal-smile:
On every major update, a snapshot of the "root" LV is taken (on SEAPATH Debian, it includes the whole rootfs, /boot included).

- Boot count is stored as a Grub environement file in ESP: /boot/efi/bootcountenv.
- The boot count is incremented each boot by `11_seapath_boot_count`.
- On successful boot (measured by system_check.service based on systemctl is-system-running), boot counting is disabled and the default boot process is used (lasted kernel first).
- On the 3rd unsuccessful boot in a row, the LVM snapshot on the root LV is recovered and the system rebooted, effectively reverting any update done since the snapshot (kernel update included).

The recovery process is implemented using initramfs scripts run by a "safe" kernel (second to last or the unique kernel).

Known limitations:
* This implementation is tied to the default configuration (disk partitioning, LV naming, ...)
* The recovery won't work if the "safe" kernel refuses to boot.


ycongal-smile:
FWIW, I tested the new ansible role integration on a SEAPATH standalone VM:
- [x] Installation
- [x] Recovery (caused by a malfunctioning userland service)
- [x] Revert

dupremathieu (commented):
Would it be possible to manage the boot/update/rollback in a more standardized way? We could rely on BLS (https://uapi-group.org/specifications/specs/boot_loader_specification) for the boot and OStree for the updates, for instance. Maybe we could discuss it in the next maintainer meeting.

dupremathieu (approved):

PR #741, opened 2025-02-26
Ci fix

yairpod:
This PR includes a list of fixes found to break the CI tests on CentOS and the CI test code for CentOS.

### DIFF
--- a/roles/ci_centos/files/cukinia.conf
+++ b/roles/ci_centos/files/cukinia.conf
@@ -1,0 +1,126 @@
+# Copyright (C) 2025 Red Hat, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+
+
+# --- Ceph tests ---
+cukinia_log "$(_colorize yellow "--- check Ceph status ---")"
+
+_ceph_status=$(timeout 10s ceph status 2>/dev/null)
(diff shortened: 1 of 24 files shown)


roles/ci_centos/files/cukinia.conf
```diff
@@ -0,0 +1,124 @@
+# Copyright (C) 2025 Red Hat, Inc.
```
dupremathieu, line 1: Would it be possible to move this file in the https://github.com/seapath/cukinia-tests/ repo?

dupremathieu (changes requested):
Some of your commit's titles are too long and GitHub cut them.
e.g.
```
 Do not skip_reboot_setup_network on CentOS as the NetworkManager need…
…s to be replaced with systemd-networkd
```

roles/ci_centos/files/cukinia.conf
yairpod: Maybe, but it's entire purpose is to create a separate test list compatible with CentOS,
How would we maintain this separation in this repo?

dupremathieu (dismissed):

dupremathieu (approved):

PR #747, opened 2025-03-09
remove backup-restore and adapt vmmgrapi on standalone

insatomcat:
Remove backup-restore on standalone
On standalone backup-restore does not make sense.

----
vmmgrapi: refactor to make it work on standalone
- admin_cluster_ip is replaced by a role variable vmmgrapi_ipaddress, specific to the vmmgrapi role
- deployment of this role is delayed to after configure_ha
- the vmmgrapi role now includes the logic to change the pacemaker configuration, instead of configure_ha

### DIFF
--- a/roles/vmmgrapi/tasks/main.yml
+++ b/roles/vmmgrapi/tasks/main.yml
@@ -136,3 +136,21 @@
     enabled: no
   when:
     - services['gunicorn.service'] is defined
+
+- name: Run extra CRM configuration commands for vm-mgr http api
+  command:
+    cmd: crm -d config load update -
+    stdin: "{{ vmmgrapi_cmd_list }}"
+  when:
+    - enable_vmmgr_http_api is defined
+    - enable_vmmgr_http_api is true
+    - "'cluster_machines' in group_names"
+  run_once: true
+  register: vmmgrapi_cmd_list_task
+  changed_when: "'CIB commit successful' in vmmgrapi_cmd_list_task.stdout"
+  vars:
+    vmmgrapi_cmd_list: |
+      primitive ClusterIP IPaddr2 params ip={{ vmmgrapi_ipaddress }} cidr_netmask=32 op monitor interval=30s meta target-role=Started
+      primitive vmmgrapi systemd:nginx.service  op monitor interval=30s
+      colocation vmmgrapi_colocation inf: ClusterIP vmmgrapi
+      order order_ClusterIP_vmmgrapi ClusterIP vmmgrapi
--- a/playbooks/seapath_setup_main.yaml
+++ b/playbooks/seapath_setup_main.yaml
@@ -46,6 +46,8 @@
   import_playbook: cluster_setup_users.yaml
 - name: Import cluster_setup_ha playbook
   import_playbook: cluster_setup_ha.yaml
+- name: Import seapath_setup_vmmgrapi playbook
+  import_playbook: seapath_setup_vmmgrapi.yaml
 
 - name: Restart all hosts
   hosts:
--- a/playbooks/seapath_setup_prerequisdebian.yaml
+++ b/playbooks/seapath_setup_prerequisdebian.yaml
@@ -17,7 +17,6 @@
   become: true
   roles:
     - debian_physical_machine
-    - vmmgrapi
 - name: Backup_restore on cluster machines
   hosts:
     - cluster_machines
--- a/playbooks/seapath_setup_prerequisdebian.yaml
+++ b/playbooks/seapath_setup_prerequisdebian.yaml
@@ -17,8 +17,13 @@
   become: true
   roles:
     - debian_physical_machine
-    - backup_restore
     - vmmgrapi
+- name: Backup_restore on cluster machines
+  hosts:
+    - cluster_machines
+  become: true
+  roles:
+    - backup_restore
 - name: Prerequis hypervisor debian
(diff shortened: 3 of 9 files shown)


eroussy (changes requested):
vm-manager can be used in standalone.
It is actually deployed on Debian standalone machines in the role ` debian_physical_machine` and on Yocto standalone machines in meta-seapath.
I think we should keep backup restore and vmmgrapi on standalone.

insatomcat:
This is not about vm-manager but about the http api.  Right now it listens on "admin_cluster_ip" which makes no sense for standalone https://github.com/seapath/ansible/blob/97d0dc1ac3f86f8812652decb2d91b1575b42b8a/roles/vmmgrapi/templates/nginx.conf.j2#L14
 backup-restore is about exporting the rbd to qcow2, it has nothing to do on a standalone.

eroussy:
> This is not about vm-manager but about the http api. Right now it listens on "admin_cluster_ip" which makes no sense for standalone
> 
> https://github.com/seapath/ansible/blob/97d0dc1ac3f86f8812652decb2d91b1575b42b8a/roles/vmmgrapi/templates/nginx.conf.j2#L14
> 
> backup-restore is about exporting the ceph rbd to qcow2, it has nothing to do on a standalone.

Agreed for backup restore
But I don't understand why we should disable vmmgrapi.
I should also work on standalone, so may be we have to rename this `admin_cluster_ip` variable ?

insatomcat:
ok done

eroussy (changes requested):
Looks good to me.
Just some details : 
- Put the default value in a `default` folder in `vmmgrapi` role, to be consistent with other roles.
- Update the `roles/vmmgrapi/README.md` file with default value for `admin_cluster_ip`
- Typo in your commit message title "backup-restorei"

insatomcat:
>     * Typo in your commit message title "backup-restorei"

OK

>     * Put the default value in a `default` folder in `vmmgrapi` role, to be consistent with other roles.

admin_cluster_ip and ansible_host are not role variables, they are global variables (admin_cluster_ip is also used in configure_ha for instance), we can't set their default inside the role.

>     * Update the `roles/vmmgrapi/README.md` file with default value for `admin_cluster_ip`

I will remove admin_cluster_ip from this list.

eroussy:
> admin_cluster_ip and ansible_host are not role variables, they are global variables (admin_cluster_ip is also used in configure_ha for instance), we can't set their default inside the role.

Oh yeah, I didn't see that.
But now we have a problem : `admin_cluster_ip` variable is not documented anywhere.

We have two options : 
- Consider this variable as a global variable and document it in the inventory examples
- Consider this variable as two role variables with the same name : One for vmmgrapi and one for configurer_ha.

The second option is the best IMO, so we should still keep the variable in the README and the default value in a default folder.

What do you think ?

eroussy:
After all, I don't really understand why the "extra CRM configuration commands for vm-mgr http api" is in the configure_ha role ?
It would be better to put that inside the vmmgrapi role ? or in a dedicated role ?

insatomcat:
> After all, I don't really understand why the "extra CRM configuration commands for vm-mgr http api" is in the configure_ha role ? It would be better to put that inside the vmmgrapi role ? or in a dedicated role ?

I agree it's related to vmmgrapi but when we deploy vmmgrapi (in the prerequisites playbook), pacemaker is not installed yet... + there is no pacemaker at all in standalone 
This leads to reordering the deployment of roles in the playbooks, and to limit the "extra crm" part to cluster_machines.

eroussy:
> > After all, I don't really understand why the "extra CRM configuration commands for vm-mgr http api" is in the configure_ha role ? It would be better to put that inside the vmmgrapi role ? or in a dedicated role ?
> 
> I agree it's related to vmmgrapi but when we deploy vmmgrapi (in the prerequisites playbook), pacemaker is not installed yet... + there is no pacemaker at all in standalone This leads to reordering the deployment of roles in the playbooks, and to limit the "extra crm" part to cluster_machines.

I see. This is a much larger work todo.
May be you can open an issue on that ?

insatomcat:
please take a look at what I just pushed.

vmmgrapi: refactor to make it work on standalone:
- admin_cluster_ip is replace by a role variable vmmgrapi_ipaddress, specific to the vmmgrapi roles
- deployment of this role is delayed to after configure_ha
- the vmmgrapi role now includes the logic to change the pacemaker configuration, instead of configure_ha

roles/vmmgrapi/tasks/main.yml
```diff
@@ -136,3 +136,21 @@
     enabled: no
   when:
     - services['gunicorn.service'] is defined
+
+- name: Run extra CRM configuration commands for vm-mgr http api
+  command:
+    cmd: crm -d config load update -
+    stdin: "{{ vmmgrapi_cmd_list }}"
```
eroussy, line 143: I think this variable was not documented properly. Can you add it to the vmmgrapi role README ?

eroussy (changes requested):
Looks great.
Just one variable that we forget to document and we are good to go.

roles/vmmgrapi/tasks/main.yml
insatomcat: this is a local variable (to the task), it should not be exposed to the user

eroussy (approved):

PR #755, opened 2025-03-24
ci-yocto: migrate latency tests to svtrace-ansible

Paullgk:
Currently, IEC61850 latency tests are run using the playbook `ci_latency_tests`. In order to unify the execution of IEC61850 latency tests, this commit makes the needed modification to use svtrace-ansible project to execute these tests.
Only the publisher PTP synchronization is done in the SEAPATH Ansible repository.


playbooks/ci_prepare_publisher.yaml
```diff
@@ -0,0 +1,47 @@
+# Copyright (C) 2024 Savoir-faire Linux, Inc.
```
eroussy, line 1: Out of curiosity : Why do the PTP synchronisation of the publisher in ansible and not directly in svtrace ?
Otherwise looks good to me.

eroussy: Nevermind, I found the answer on svtrace :
> as now PTP is a requirement to use svtrace-ansible

eroussy (dismissed):

Paullgk:
This PR must be merged to fix Yocto CI, and enter in conflict with #801 which solve Oracle Linux and Debian CI.
As this PR is only related to Yocto, this PR is forced merge as https://github.com/seapath/ansible/actions/runs/16214877555/job/45781923117 is OK. 
Approved by @eroussy

PR #757, opened 2025-03-25
debian: adapt tests to incoming trixie

insatomcat:


roles/debian_tests/cukinia-tests/cukinia/common_security_tests.d/apt.conf.j2
```diff
@@ -13,6 +15,25 @@ id "SEAPATH-00197" as "Only official apt repositories configured: {{ repo }}" \
 {% endfor %}
 {% endif %}
 
+{% endif %}
+
+{% if ansible_distribution == 'Debian' and ansible_distribution_release == 'trixie' %}
+
+{% if apt_repo is defined %}
+id "SEAPATH-00197" as "Only wanted apt repositories configured: only use sources.list.d" \
+    cukinia_test -z "$(ls /etc/apt/sources.list)"
```
dupremathieu, line 24: I think you want to use `cat` not `ls`?

dupremathieu (changes requested):

roles/debian_tests/cukinia-tests/cukinia/common_security_tests.d/apt.conf.j2
insatomcat: The point here is to check this file does not exist, so either ls or cat would work, the check is to verify the resulting string is null.
ls seems lighter.

dupremathieu: Ha, ok, I've thought you wanted to test if the file is empty.
In this case, the simplest way is `cukinia_test ! -e /etc/apt/sources.list`.

insatomcat: thx, done

dupremathieu (approved):

PR #802, opened 2025-07-08
Network interface aliases for Welotec RSAPCMK2

Paullgk:
Hello,
This (draft) PR makes the needed changes for creating the needed network interface aliases for  Welotec RSAPCMK2 hypervisor. This work is based on [Welotec build_debian_iso fork], and reproduce the same behavior.
Currently, the implementation is voluntary made in a playbook, and can't be merged as it as these modification are for a specific hypervisor hardware, and we currently don't have any support for this in Ansible.

Feel free to give me your opinion!

[Welotec build_debian_iso fork]: https://github.com/welotec/seapath-build_debian_iso/blob/main/usercustomization/scripts/USERCUSTOMIZATION/90-welotec

### DIFF
--- a/playbooks/seapath_setup_custom_hardware.yaml
+++ b/playbooks/seapath_setup_custom_hardware.yaml
@@ -1,0 +1,11 @@
+
+---
+- name: Welotec hardware customization
+  hosts:
+    - cluster_machines
+    - standalone_machine
+  become: true
+  gather_facts: true
+  roles:
+    - role: hardware_customization/welotec
+      when: welotec is defined and welotec | bool
--- a/roles/hardware_customization/welotec/README.md
+++ b/roles/hardware_customization/welotec/README.md
@@ -1,0 +1,27 @@
+# Welotec hardware customization role
+
+This role applies specific configuration for Welotec hardware:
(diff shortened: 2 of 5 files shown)


Paullgk:
@dknueppe

Paullgk:
> Hey Paul, thanks for looking into it and taking the time. It looks like in the mac_address_map there is lan8 missing. Also my script made sure it found 8 consecutive MACs with Welotecs OUI. In case there ever are other NICs with Welotecs OUI that was a way to ensure I found the onboard (on the mainboard) NICs. though that is at the moment not really necessary.
> 
> I also wanted to add that the Alternative names aren't strictly needed. The idea back then was that I could set them at whatever point I'd like (during or after installation) and would target those in the inventory files for all the further things - it seemed like a sensible abstraction so that I could reuse inventory files without having to seperately modify according to how I wired up the network. So I could target lan-ptp or lan-mgmt as opposed to enp2s0 and enp4s0 or whatever.

Thanks for the feedback! I think @insatomcat recommendation is nice, but we still need some tasks to map the appropriate MAC address to the correct NIC. 
The tasks I have implemented also check Welotec OUI!

PR #818, opened 2025-09-19
bump cukinia and vm_manager

insatomcat:


roles/network_buildhosts/tasks/main.yml
```diff
@@ -9,8 +9,8 @@
 - name: Build hosts file (Cluster Network)
   lineinfile:
     dest: /etc/hosts
-    regexp: '.*{{ hostvars[item].hostname | default(item) }}(| rbd)$'
-    line: "{{ hostvars[item].cluster_ip_addr }} {{ item }} {{ hostvars[item].hostname | default('') }}{% if hostvars[item].hostname is defined %} {% endif %}{% if item == inventory_hostname %}rbd{% endif %}"
+    regexp: '.*{{ hostvars[item].hostname | default(item) }}$'
+    line: "{{ hostvars[item].cluster_ip_addr }} {{ item }} {{ hostvars[item].hostname | default('') }}"
```
eroussy, line 13: This means that the `hostname` variable is now required
Please update the README of the role and make it fallback to inventory hostname in the default

dupremathieu: No, there is a default item.

PR #822, opened 2025-09-24
Containers

bkkouame:


ra/docker
```diff
@@ -0,0 +1,292 @@
+#!/bin/sh
```
dupremathieu, line 1: You can remove this commit

dupremathieu (changes requested):
Remove the Change-ID.
The commit wrapping is wrong.
Add a PR description.

dupremathieu:
> Seapath is moving to cephadm, and cephadm defaults to podman containers. I feel the target for running containers on seapath should be podman. I for one am not confortable having podman containers for ceph and docker daemon and containers running at the same time on the same host.

Agree. This PR should be completly rework.

dupremathieu:
This PR should be completely reworked.
I have converted to draft.

PR #830, opened 2025-11-25
fix root ssh key generate hung issue

ggoklani:
While running main Seapath setup playbook on Centos stream 9 , root ssh key generate task is getting hung this task is part of playbooks/cluster_setup_users.yaml.

RCA :
Current code uses ansible user module with generate_ssh_key

Module Internal Behavior: The user module's generate_ssh_key feature internally calls ssh-keygen, but the way it handles this might not be fully non-interactive in all edge cases or Linux distributions, especially with system accounts.

Solution : Used shell approach
 The shell approach gives you complete control and transparency, avoiding the "black box" behavior that caused the hang with the user module.


insatomcat:
I had this problem when working with debian13 and ansible 2.16. This is how I patched it, hoping we could soon use a patch ansible version:
https://github.com/seapath/ansible/commit/d2aaae78cd246bd523169f95b6883cb8140b959c

eroussy:
> @eroussy Thanks for your Update.. Issue has been resolved after using ansible version 2.16 with PR #843

Good to know
We will probably close this PR when #843 is merged

PR #836, opened 2025-12-03
CentOS SELinux fixes (ptpvsock, ptp4l) and hardening profile fix

matosatti:
Allow ptpvsock and ptp4l to execute (by adding SELinux rules), and
fix the debian hardening profile addition of sshd wait for DNS.


roles/debian_hardening/tasks/main.yml
```diff
@@ -167,11 +167,23 @@
     state: absent
   when: revert
 
+- name: Set default SSH service file path
+  set_fact:
+    ssh_service_path: "/lib/systemd/system/ssh.service"
+    ssh_service_newline: "After=network.target auditd.service network-online.target"
+  when: seapath_distro != "CentOS"
```
dupremathieu: The debian_hardening role is supposed to be a Debian-only role.

You should create a centos_hardening role for CentOS.

If you think more than 80% of the code is common, maybe we can change this role to be a generic hardening role for all distros.

eroussy:
Closing this for inactivity.
Feel free to reopen if needed

PR #838, opened 2025-12-04
Adds role to deploy cephFS

insatomcat:


roles/deploy_cephfs/tasks/main.yml
```diff
@@ -0,0 +1,165 @@
...
+- name: Find the first node with an active monitor
+  set_fact:
+    first_mon_node: "{{ (ceph_status_results.results | selectattr('rc', 'equalto', 0) | map(attribute='item') | list)[0] }}"
+
+  run_once: true
+
+- name: Debug first monitor node
+  debug:
+    msg: "First monitor node detected: {{ first_mon_node }}"
+  run_once: true
```
eroussy: I would suggest adding `verbosity`. 1 at least

roles/deploy_cephfs/tasks/main.yml
```diff
@@ -0,0 +1,165 @@
...
+
+- name: Create CephFS volume 'seapathcephfs' on the first monitor
+  command: ceph fs volume create seapathcephfs
+  register: create_fs
+  changed_when: "'already exists' not in create_fs.stderr"
+  ignore_errors: yes
```
eroussy: Why this `ignore_errors` ?
If necessary, we should better use `failed_when`

roles/deploy_cephfs/tasks/main.yml
```diff
@@ -0,0 +1,165 @@
...
+    dest: /etc/ceph/volumemount.secret
+    owner: root
+    group: root
+    mode: '0600'
+
+- name: Deploy MDS for CephFS volume 'seapathcephfs'
+  command: ceph orch apply mds seapathcephfs --placement="count:3"
+  register: mds_deploy
+  changed_when: "'already exists' not in mds_deploy.stderr"
+  ignore_errors: yes
```
eroussy: why ignore_errors ?

roles/deploy_cephfs/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Requirements
+
+You need to use cephadm and not ceph-ansible to use this feature
+
+## Role Variables
+
+| Variable                      | Required | Type   | Default                                    | Comments                                                                 |
+|-------------------------------|----------|--------|--------------------------------------------|--------------------------------------------------------------------------|
+| `deploy_cephfs_localdirtoupload` | No       | string | `../inventories_private/cephfs_upload/`     | Local directory on the Ansible controller that will be synchronized to the remote CephFS mount. Can be overridden at playbook, group_vars, host_vars, or CLI. |
```
eroussy: Change comment for "Local directory that will be uploaded to ceph_fs volume once created."

roles/deploy_cephfs/files/ceph-mds-ready@seapathcephfs.service
```diff
@@ -0,0 +1,12 @@
+# Copyright (C) 2025 RTE
+# SPDX-License-Identifier: Apache-2.0
```
eroussy: I'm not sure of the purpose of this service ? Is it just to flag if ceph_fs is running correctly ?
In that case, put it in the description of the service or in the service file comments

eroussy (changes requested):
Thanks, that looks great.
I left some comments

I will try to test that on a Yocto cluster soon (probably early next week)

PR #851, opened 2026-01-12
Replace livemigration user with libvirtadmin user

### LINKED ISSUE #846: Too many users have access to the libvirt socket
**Abstract**
Multiple users are configured to have access to the Libvirt socket on Debian (admin, libvirt, livemigration)
Only one user should be configured for that purpose.

**Current status**
On Debian, three users have access to the libvirt socket:
- admin (the administration user)
- libvirt (dedicated user, but seems unused)
- livemigration (used only for livemigration)
Additionnaly, both admin and livemigration can connect from one cluster machine to another on the cluster by SSH

On Yocto, only `livemigration` can access the libvirt socket and connect between machines on the cluster.
To use vm_manager console command, one must specify the user as livemigration:
`vm-mgr console --ssh-user livemigration <vm_name>`

**Detailed Description**
Only `livemigration` user must be able to
- have access to the Libvirt socket
- connect between cluster machines
Because it will be used also for vm console access, this user should be renamed `libvirtadmin`

**Work to be done**
- rename `livemigration` to `libvirtadmin`
- remove `libvirt` user on Debian
- remove libvirt socket access for `admin`
- remove admin ssh keys on the cluster (see #842)
- select `libvirtadmin` as default user for vm console on vm_manager

eroussy:
- Rename `livemigration` user to `libvirtadmin`
- Force the creation of the `libvirtadmin` user (not conditional anymore)
- Remove the `libvirt` user on Debian
- Remove the admin user from the libvirt group
- Remove the SSH key exchange for the admin user over the cluster

Closes #846 

API break:
- `libvirtadmin` user is always created and always called `libvirtadmin`
- `livemigration_user` can still be chosen when deploying VMs (default to `libvirtadmin`)

### DIFF
--- a/playbooks/seapath_setup_libvirt.yaml
+++ b/playbooks/seapath_setup_libvirt.yaml
@@ -9,4 +9,5 @@
   hosts: hypervisors
   become: true
   roles:
+    - detect_seapath_distro
     - configure_libvirt
--- a/roles/configure_libvirt/defaults/main.yml
+++ b/roles/configure_libvirt/defaults/main.yml
@@ -2,4 +2,4 @@
 # SPDX-License-Identifier: Apache-2.0
 
 ---
-configure_libvirt_allow_non_root_libvirt_socket_access: false
+configure_libvirt_allow_non_root_libvirt_socket_access: true


roles/configure_admin_user/README.md
insatomcat: I think you forgot to remove the tasks/main.yml file (or maybe it's because of a rebase), but anyhow, the role folders and the tasks/main.yml file still exist in your resulting branch

roles/centos_physical_machine/tasks/main.yml
insatomcat: you did both debian and centos, can you include the same thing for the oraclelinux role?

insatomcat (changes requested):

eroussy:
> I'm rebasing on newer version of main, since ansible2.16 changed a lot of things and the debian CI should now be debian13.

Thanks for notifying me. But that's better to let me do the rebase in that case, because I will push changes to correct your comments.
It avoids triggering the CI two times.

( Just for information, the CI did pass on Debian 13 the first time. The rebase problem was only the commit of #855 )

insatomcat (approved):

PR #855, opened 2026-01-15
Fixed result type of "when" conditions.

dahoat-sprecher:
The "when" condition in tasks of role network_systemdnetworkd and detect_seapath_distro returned values which could be evaluated to a boolean but were no real booleans and thus rejected by Ansible (2.19.4). 
OS: Debian Trixie (6.12.63-1 (2025-12-30) x86_64)

Best regards, 
Daniel


dahoat-sprecher:
Hi @eroussy 
Thanks for notifying me. My fork should be synced ad the PR updated now.

Issue #442, opened 2024-03-20
Replace ceph-ansible by Cephadm


dupremathieu:
## Abstract

Cephadm is tools to configure and deploy SEAPATH: https://docs.ceph.com/en/quincy/cephadm/.
Cephadm relies on containers. With the cephadm-ansible collection, it replaces ceph-ansible.

Note that ceph-ansible we actually used to configure Ceph will no longer get new features and will not support the future Ceph version (but it is still maintained).

## Current status

* Ceph is installed directly inside SEAPATH images.
* Ceph version differs between SEAPATH variants.
* In both variants, the Ceph version is quite old and will be deprecated soon.
* We use ceph-ansible to setup Ceph
* ceph-ansible version differs between SEAPATH variants (because it has to follow the Ceph version).
* ceph-ansible requires an old version of Ansible to be used.

## Detailed Description

The biggest change of cephadm-ansible regarding ceph-ansible is cephadm only supports containers. The installation will be done in two steps:

* first, we install ceph only on one machine of the cluster
* after cephadm extends the installation to other machines We also need to setup a registry to store Ceph images containers. This registry can be one machine of the cluster or inside an external machine. It can be removed after the installation.

## Benefit to SEAPATH

* This is the recommended way to configure and deploy Ceph.
* The Ceph version will no longer be linked to the distribution.
* The security will be enhanced with the container isolation used by Cephadm.
* Greatly reduce the Yocto variant generation times.
* The Ansible version will be no longer restricted by ceph-ansible
* Easier Ceph update which is now managed by Cephadm

## Works to be done

* Remove Ceph from SEAPATH images.
* Modify our Ansible files to use Cephadm-ansible instead of ceph-ansible.
* Use Ansible to create a local registry to push Ceph container images on it to allow offline installation

insatomcat:
Very interesting and we will definitely have to work on this at some point.
I have 2 questions:

> Modify our Ansible files to use Cephadm-ansible instead of ceph-ansible

The description of the cephadm-ansible project states "[cephadm](https://docs.ceph.com/en/latest/cephadm/)-ansible is a collection of Ansible playbooks to simplify workflows that are **not** covered by cephadm"
Not quite sure what it means, but it looks like it's quite limited, and does not help create the cluster for example.

> Use Ansible to create a local registry to push Ceph container images on it to allow offline installation

In our current philosophy, the build of the installation media is the only time we need to be online. 
The ansible part does current not require an internet connection.
[...]

PR #867, opened 2026-02-18
feat: replace raw cephadm commands with ceph.cephadm collection modules

fabiendupont:
Replace raw `command:` calls to cephadm/ceph CLI with idempotent modules from the ceph.cephadm Ansible collection across day-1 (deploy) and day-2 (remove/replace) operations.
    
Key changes:
 - Add ceph.cephadm collection dependency (ansible-requirements.yaml)
 - Use cephadm_bootstrap, cephadm_registry_login, ceph_orch_host, ceph_orch_apply, and ceph_config modules in place of command: tasks
 - Simplify SSH key distribution using ceph cephadm get-pub-key
 - Split multi-document spec.yaml.j2 into single-document templates for ceph_orch_apply compatibility
 - Add cephadm_preflight.yaml for host preparation (Debian, CentOS, OracleLinux, Yocto)
 - Rewrite machine removal with graceful drain before host removal, falling back to forced offline removal when host is unreachable
 - Add complete machine replacement playbook (drain, remove, add, OSD)
 - Extract setup_user.yml for independent inclusion from playbooks
 - Fix ceph.conf.j2 bug: replace undefined cephadm_network variable with public_network / cluster_network from inventory
 - Apply ceph_conf_overrides via ceph_config (was silently ignored)

Depends on #866


insatomcat:
Thanks for this work, and sorry it sat unreviewed for so long. Closing it, with the reasoning here so nothing is lost.

**What changed on main.** The merge base of this branch is from February. Since then the `cephadm` role was split into `cephadm_install` and `cephadm`, the distro roles were consolidated into generic `configure_seapath_distro`, `configure_physical_machine` and `configure_hypervisor` roles that also cover SLES, and the `replace_machine_*_cephadm.yaml` playbooks were replaced by `cluster_remove_machine.yaml`. About 86 files and 1700 lines moved in the area this PR touches, and #442 was closed in July. Rebasing this branch would mean rewriting it.

**On the `ceph.cephadm` dependency.** The collection is absent from Galaxy, so it can only be installed from git, and this PR pins `devel`, a moving branch: any upstream commit lands straight in our CI and builds stop being reproducible. Our one git-sourced dependency, `openstack.config_template`, is pinned to the release tag `2.1.1`, and even so we had to move it from opendev to the GitHub mirror in August after repeated outages broke the CI, so the cost of hosting a dependency outside Galaxy is something we have already paid. A git dependency is also a problem for the collection we publish on Galaxy. The collection ships seven modules, and OSD lifecycle, zapping, pools and CephX are outside them, so the role stays a hybrid either way. The PR also has to split `spec.yaml.j2` into single-document templates and write its own preflight for our distros. What remains is idempotency and check mode on bootstrap and host add, which does not pay for that.

**What we kept.** Three ideas here were worth having on their own:

- applying `ceph_conf_overrides` on a running cluster: #1014
- the graceful side of machine removal, as a guard that refuses to force-remove a machine still serving OSDs: #1015
- separate `public_network` and `cluster_network`: dropped on purpose. On a three-node hyperconverged cluster both networks share the same link unless a NIC is dedicated to replication, so the split buys no isolation while adding a failure mode where OSDs flap when the cluster network degrades on its own and the mons stay reachable. One correction while we are here: `cephadm_network` is defined in the role README and in the example inventory, so it was not the undefined variable the description mentions.

The registry role of #866 is a separate discussion, see the comment there.

PR #875, opened 2026-02-23
[Proposal] Add ansible molecule to test roles

AntoineDupre:
Here a proposal to add [ansible molecule](https://docs.ansible.com/projects/molecule/) in seapth ansible.
This PR introduces Molecule into the repository in order to add unit testing for Ansible roles.

## Why molecule
Currently, roles are tested through full playbook execution as part of the global deployment workflow.
While this validates integration at the playbook level, it does not provide:
- Isolated unit testing of individual roles
- A simple and lightweight test environment for contributors
- A fast feedback loop when modifying a single role
- Automated validation of Ansible best practices (such as idempotency)

Molecule addresses these limitations by allowing each role to be tested independently from the full Seapath deployment environment. It brings:
- Independent testing of roles (unit-style testing)
- A reproducible and disposable test environment
- A contributor-friendly workflow (`molecule test` inside a role)

### How Molecule works
Molecule orchestrates a full test around an Ansible role.
A role test is organized into scenarios, located under:
```
molecule/<scenario_name>/
```
Each scenario contains a `molecule.yml` file describing the test configuration.
A scenario typically includes:
- [driver](https://docs.ansible.com/projects/molecule/pre-ansible-native/?h=drivers#driver): Defines how the test environment is created (Docker, Podman, delegated, etc.).
- [provider](https://docs.ansible.com/projects/molecule/pre-ansible-native/?h=drivers#provisioner): Defines how the role is executed. The standard and recommended method is to use Ansible itself.
- [verifier](https://docs.ansible.com/projects/molecule/pre-ansible-native/?h=drivers#verifier):  Defines how validation is performed after role execution. Verification can be implemented using Ansible assertions in a playbook, but Molecule also supports tools such as `testinfra`.

For each scenario, Molecule runs test sequence like:
```
create → prepare → converge → idempotence → verify → destroy
```
By default, Molecule runs the role twice to validate idempotency (i.e., the second run should not report changes).

It is important to note that Molecule does not introduce new testing logic that could not be achieved with Ansible alone.
It provides a structured and standardized framework around Ansible to:
- Create test environments
- Execute roles
- Validate results
- Assert code coverage
- Enforce best practices

## What this PR introduces

This PR provides an initial implementation of Molecule tests for the `network_basic` role.
It demonstrates:
- How to structure Molecule scenarios
- How to use Ansible as both provider and verifier
- How to test different behavioral paths of a role

Two scenarios are provided:
**1. `default` scenario**
Tests the standard behavior of the role.
It includes:
`molecule.yml`: Defines the test environment configuration.
`converge.yml`: A playbook that applies the network_basic role with default variables.
`prepare.yml`: Sets up the test environment before the role execution.
`verify.yml`: Verifies that the expected network configuration is present using Ansible assertions.

This scenario validates the normal usage of the role.

**2. `wipe` scenario**
Tests the behavior of the role when using the variables that wipe network directories before deployment.

This scenario specifically validates that:
- Existing network configuration directories are removed
- The expected configuration is re-deployed correctly
 
**idempotency disabled for this scenario**
By default, Molecule runs an idempotency test, meaning it executes the role twice to ensure that the second run produces no changes.
However, in this case, the role:
- Removes all elements from the target directory
- Then re-deploys configuration files

This workflow is intentionally destructive and cannot be strictly idempotent in the classical sense.

For this reason, the idempotency step has been disabled for this scenario.

## Addition of Tox for test orchestration

This PR also introduces [tox](https://github.com/tox-dev/tox) at the repository root. It brings:
- Provides a reproducible Python virtual environment for running tests
- Avoids requiring contributors to install Ansible and Molecule globally
- Centralizes test dependencies
- Makes it possible to test against different Ansible versions if needed

Molecule test can be directly run from each role directory:
```bash
cd roles/network_basics
molecule test --scenario-name default  --report --command-borders
```

It requires some orchestration to be run from the seapath-ansible root directory. Tox provides this orchestration.
Tests can now be executed from the repository root with:
```bash
tox -e molecule
```
This command:
-Discovers roles that contain a molecule/ directory
- Runs `molecule test --all` for each of them

## Conclusion
The goal of introducing Molecule in this repository is to progressively generalize the practice of adding unit tests for Ansible roles, especially for the development of future roles.

This initial implementation should be seen as a foundation rather than a final state. Over time, we aim to:
- Extend Molecule coverage to the existing code base
- Improve overall quality by making it easier to test specific configurations in isolation
- Detect issues in parts of the code that are currently not exercised by the CI pipeline
- Gain better visibility into test coverage at the role level
- Identify roles that may need refactoring to become more easily testable

Introducing systematic role testing will also help highlight roles that are not strictly idempotent.
In some cases, this may encourage restructuring them to better align with Ansible’s philosophy: describing a desired state rather than executing a sequence of imperative actions.

This PR is a first step toward a more test-driven and quality-oriented approach to role development within the project.

### DIFF
--- a/molecule_requirements.txt
+++ b/molecule_requirements.txt
@@ -1,0 +1,3 @@
+ansible-core
+molecule
+molecule-plugins[podman]
--- a/tox.ini
+++ b/tox.ini
@@ -5,9 +5,7 @@ skipsdist = true
 [testenv:molecule]
 description = Run Molecule tests for all roles that have molecule/<scenario>/
 deps =
-    ansible-core
-    molecule
-    molecule-plugins[podman]
+    -r molecule_requirements.txt
 allowlist_externals =
     bash
 commands =


roles/network_basics/molecule/default/molecule.yml
```diff
@@ -0,0 +1,24 @@
+---
+dependency:
+  name: galaxy
+
+driver:
+  name: docker
```
dupremathieu: Is it possible to use the Podman driver instead of Docker?

Could you add a requirements.txt?

dupremathieu (approved):

PR #879, opened 2026-02-27
Add missing exit 1

AntoineDupre:
`prepare.sh` misses `exit 1` if ansible version is different that 2.16

also, README seems outdated:
```
Without `cqfd` you need to install the dependencies manually.
The client machine that is going to run Ansible must have Ansible 2.10 installed
an Inventory file and playbook files to play. To install Ansible 2.10 on this
machine please refer to the Ansible documentation at

*Warning:* Currently only the Ansible version 2.10 is supported. Other versions
will not work.
```

Shall it be update to 2.16 ?


eroussy:
Yes, the readme should be updated
Good catch

PR #883, opened 2026-03-02
[Fix] Remove docker service tasks

AntoineDupre:
`Enable docker service` tasks does nothing on roles:
- roles/debian_physical_machine/tasks/main.yml
- roles/centos_physical_machine/tasks/main.yml

It misses `enable: true` and `state: started`

### DIFF
--- a/roles/centos_physical_machine/tasks/main.yml
+++ b/roles/centos_physical_machine/tasks/main.yml
@@ -145,9 +145,6 @@
     enabled: yes
     state: started
 
-- name: Enable docker.service
-  ansible.builtin.systemd:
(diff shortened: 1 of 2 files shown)


insatomcat:
not quite sure about this one for 2 reasons:

1. docker was (historically) only needed for the ptpstatus service, but it's not true anymore, docker is currently not needed to what I can recall
2. even for upcoming container needs (I'm thinking prometheus exporter for instance), we are most likely transitionning to podman https://github.com/seapath/TSC/issues/43

insatomcat:
yes

PR #892, opened 2026-03-06
VNC console for VMs

dahoat-sprecher:
Hello to All, 

During deployment of Windows VMs, we encountered issues requiring manual intervention. Therefore, we added a VM feature for a VNC based console. By default, it listens only on 127.0.0.1 and is accessible using port forwarding (e.g. `ssh -L 5900:localhost:5900 node1`).

Best regards, 
Daniel


dahoat-sprecher:
Documentation is added, sorry for overlooking this in the first place.

PR #893, opened 2026-03-06
Additional ceph disks for clusters

dahoat-sprecher:
Hello to All, 

As we noticed, the guest.xml.j2 contained partial support for (one) additional disk(s) for VMs. This Pull Request adds this feature also to clusters for an arbitrary amount of disks. However, this contains also a breaking change as the inventory for a VM uses a map instead of a list: 
```
 additional_disk:
     vdb: "../files/data.qcow2" # TODO: Replace with your disk image
```
However, this permits explicitly naming additional disks. 
See also: https://github.com/seapath/vm_manager/pull/84

Best regards, 
Daniel

### DIFF
--- a/library/cluster_vm.py
+++ b/library/cluster_vm.py
@@ -201,6 +201,13 @@ options:
       - Optional parameter relevant only if I(command) is C(clone)
     type: bool
     default: false
+  additional_disks:
+    description:
+      - List of qcow2 file paths for additional disks
+      - Device names (vdb, vdc, ...) are assigned automatically
+      - Optional parameter relevant only if I(command) is C(create)
+    type: list
+    default: []
   disk_bus:
     description:
       - Disk bus type to use for the VM's disk (virtio, scsi, ide, etc.)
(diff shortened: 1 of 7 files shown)


library/cluster_vm.py
```diff
@@ -201,6 +201,14 @@
       - Optional parameter relevant only if I(command) is C(clone)
     type: bool
     default: false
+  additional_disks:
+    description:
+      - Dictionary mapping device names to qcow2 file paths for additional disks
```
eroussy: Is it necessary to have a dictionary ?
The current additional_disk variable is a list of strings. You could keep the same type to avoid breaking change

inventories/examples/seapath-vm-deployement.yaml
```diff
@@ -43,6 +43,10 @@ VMs:
       # Configuration part. These variables only work with the guest.xml.j2 template
       vm_features: ["secure-boot"] # features for non real-time VM with secure boot enabled
       nb_cpu: 1 # TODO : Number of vCPU of your VM
+      # Additional disks (optional). Keys are device names, values are qcow2
+      # file paths. In cluster mode, imported into Ceph as data_<vmname>_<dev>.
+      additional_disk:
+        vdb: "../files/data.qcow2" # TODO: Replace with your disk image
```
eroussy: Remove this.
We try to keep the example inventory short. The complete documentation should only be in the roles README

eroussy (changes requested):

dupremathieu (dismissed):

eroussy (dismissed):

dupremathieu (approved):

PR #894, opened 2026-03-06
[Proposal] Add Molecule tests for playbooks

AntoineDupre:
This PR proposes adding **Molecule tests for playbooks**, in addition to the existing Molecule tests already present for roles.

While the project already validates roles individually, testing playbooks themselves is also important because they validate the integration between roles. Playbooks orchestrate roles and often introduce:

- complex variable combinations
- variable overrides between roles
- conditional logic
- dependencies between roles

Testing playbooks helps ensure that the links between roles and variables defined across different roles behave correctly when executed together.

#### Goal of this PR

The main goal of this PR is to introduce the testing logic and CI infrastructure required to test playbooks with Molecule.

For this reason, the PR includes a first playbook test used as a reference implementation, whose purpose is mainly to:

- introduce the testing approach
- validate the CI integration
- prepare the repository structure for future playbook tests

This establishes the foundation so additional playbooks can easily be added to the test suite later.

#### Changes in this PR

- Add Molecule scenarios for playbooks
- Add one initial playbook test to demonstrate the approach
- Integrate playbook testing in CI, running in parallel with existing role tests
- Update `tox` configuration to support these additional tests

#### Testing approach

The goal of the playbook scenarios is not to duplicate unit-style validation already covered by role tests. Instead, they focus on:

- validating playbook execution
- ensuring roles integrate correctly
- verifying variables and orchestration logic across roles

Special care was taken to ensure that `verify.yml` in playbook scenarios does not overlap with role-level tests. The emphasis is on executing the playbooks and validating the integration behavior, rather than re-testing internal role logic.

Note: in this initial implementation, the `verify.yml` checks cover several aspects that ideally should be validated directly in role-level tests in the future. However, since some roles currently do not yet have dedicated tests, these checks are temporarily included here to ensure a minimal level of validation.

### DIFF
--- a/.github/workflows/ci-molecule.yml
+++ b/.github/workflows/ci-molecule.yml
@@ -11,8 +11,8 @@ permissions:
   contents: read
 
 jobs:
-  molecule:
-    name: Molecule Tests
+  molecule-role:
+    name: Molecule Role Tests
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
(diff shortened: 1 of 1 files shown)


playbooks/molecule/cluster_setup_users/molecule.yml
```diff
@@ -0,0 +1,46 @@
...
+    verify: verify.yml
+
+scenario:
+  test_sequence:
+    - dependency
+    - cleanup
+    - destroy
+    - syntax
+    - create
+    - prepare
```
eroussy: We should be able to add idempotency (so enable all the check) to this list, after merging #891

PR #906, opened 2026-03-27
roles/ptp_status_vsock: PTP domain number

jkarrenpalo-abb:
Allow setting the PTP domain number that is used with the pmc command.

When using the pmc tool to read the PTP status from ptp4l, the domain numbers
used in both applications need to match.

This is related to the additions made in #902.

### DIFF
--- a/inventories/examples/seapath-cluster.yaml
+++ b/inventories/examples/seapath-cluster.yaml
@@ -38,6 +38,9 @@ all:
     subnet: 24 # Subnet mask in CIDR notation. TODO
     ntp_servers: # NTP servers IP. TODO
       - "185.254.101.25" # public NTP server example
+    ptp_domain_number: 0 # PTP domain number (0 to 255)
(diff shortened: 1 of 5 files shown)


inventories/examples/seapath-cluster.yaml
```diff
@@ -38,6 +38,9 @@ all:
     subnet: 24 # Subnet mask in CIDR notation. TODO
     ntp_servers: # NTP servers IP. TODO
       - "185.254.101.25" # public NTP server example
+    ptp_domain_number: 0 # PTP domain number (0 to 255)
+    timemaster_ptp_domain_number: {{ ptp_domain_number }}
+    ptp_status_vsock_domain_number: {{ ptp_domain_number }}
```
eroussy: Oh, sorry I forget that
ansible values must be enclosed with quotes 

```
timemaster_ptp_domain_number: "{{ ptp_domain_number }}"
ptp_status_vsock_domain_number: "{{ ptp_domain_number }}"
```
That's why the QA is complaining

PR #914, opened 2026-04-20
feature/networkmanager

meritissimo1:
Hi everyone! I'm opening this PR as a draft first just to run the CI and validate my changes.

Basically, these changes involve replacing the systemd-networkd package with NetworkManager only in Red Hat family distributions, as it is the standard for CentOS/RHEL. To make this work, I created a new dedicated role called network_networkmanager that uses nmcli to configure the management bridge and the OVS cluster interface.

I also tweaked the main network playbook so it automatically skips the Debian specific roles, like systemd-networkd and networkdwait, when running on RedHat, applying the new NetworkManager role instead.

I'll let the CI do its thing now to see how it goes :)

Cheers!

### DIFF
--- a/playbooks/seapath_setup_network.yaml
+++ b/playbooks/seapath_setup_network.yaml
@@ -14,11 +14,15 @@
   roles:
     - detect_seapath_distro
     - role: network_basics
+      when: ansible_os_family != 'RedHat'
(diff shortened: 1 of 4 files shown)


eroussy:
Closing this for inactivity.
Feel free to re-open if needed

PR #917, opened 2026-04-26
cephadm: create RBD/cephfs pool with fixed pg_num to avoid failure on small disks

dupremathieu:
When creating the RBD pool without explicit PG count, Ceph's pg generator tries to autoscale immediately. On small disks this can fail because the initial PG count is too high for the available OSD space.

Create the pool with a fixed 1 PG (1 1) first, then explicitly re-enable pg_autoscale_mode afterwards. This lets the pool exist safely on small clusters while still benefiting from autoscaling as the cluster grows.

Do the same with cephfs.

### DIFF
--- a/roles/cephadm/tasks/main.yml
+++ b/roles/cephadm/tasks/main.yml
@@ -363,7 +363,7 @@
   failed_when: false
 
 - name: Create RBD pool if it doesn't exist
-  ansible.builtin.command: ceph osd pool create rbd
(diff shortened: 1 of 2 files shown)


dupremathieu:
Closing in favor of an inventory-level fix. The actual root cause is that the example inventory sets `osd_pool_default_pgp_num: 128`, which conflicts with the PG autoscaler (which starts new pools at `pg_num = 1`). Removing that override makes pool creation succeed without any workaround. See companion issue and replacement PR.

PR #920, opened 2026-05-04
Wrynose migration

Paullgk:
Hello,
This PR adds minor fixes to prepare the upcoming release of the Wrynose LTS of Yocto on the SEAPATH Yocto version.
* Run only `deploy_cephfs` on SEAPATH Debian as this role is currently not supported on SEAPATH Yocto
* Fix a failing task in with `ceph_expansion_lv` role when a cephadm installation is run on a fresh installation of a SEAPATH Yocto hypervisor
* Enable the `netplan-configure.service` systemd service when using Netplan. This service is new since Netplan 1.2.0, and is used to run all the Netplan backend operation, including the setup of the network interfaces on startup.

### DIFF
--- a/roles/network_netplan/tasks/main.yml
+++ b/roles/network_netplan/tasks/main.yml
@@ -21,6 +21,15 @@
         mode: "0600"
       with_items: "{{ netplan_configurations }}"
       notify: Apply Netplan configuration
+    - name: Check of netplan-configure service exist on the target
+      ansible.builtin.service_facts:
+    - name: Enable netplan-configure.service
+      ansible.builtin.systemd:
+        name: netplan-configure.service
+        enabled: true
+        state: restarted
+      when:
+        - '"netplan-configure.service" in ansible_facts.services'
     - name: Avoid reboot for netplan
       ansible.builtin.set_fact:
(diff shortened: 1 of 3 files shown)


roles/network_netplan/tasks/main.yml
```diff
@@ -21,6 +21,12 @@
         mode: "0600"
       with_items: "{{ netplan_configurations }}"
       notify: Apply Netplan configuration
+    - name: Enable netplan-configure.service
+      ansible.builtin.systemd:
+        name: netplan-configure.service
+        enabled: true
+        state: restarted
+      when: seapath_distro == "Yocto"
```
dupremathieu: It should be also the case for all distro.
We should remove `when: seapath_distro == "Yocto"`

dupremathieu (changes requested):

roles/network_netplan/tasks/main.yml
dupremathieu: Or even better we can check if the service is present.

Paullgk: Done

dupremathieu (approved):

PR #923, opened 2026-05-05
inventories/examples: drop osd_pool_default_pg{,p}_num overrides

### LINKED ISSUE #922: Ceph pool creation fails on small disks due to osd_pool_default_pgp_num conflicting with autoscaler
## Problem

When deploying Ceph (RBD pool via `cephadm`, CephFS pools via
`deploy_cephfs`) on a cluster with small disks, pool creation fails:

    Error ERANGE: 'pgp_num' must be greater than 0 and lower or equal
    than 'pg_num', which in this case is 1

For CephFS the failure surfaces as a generic `Error EINVAL` because
`ceph fs volume create` relies on the mgr `volumes` module to create
the metadata/data pools internally, and that module returns `None`
instead of a proper error tuple when the underlying pool creation
fails.

## Root cause

The SEAPATH example inventory `inventories/examples/seapath-cluster.yaml`
sets in `ceph_conf_overrides.global`:

    osd_pool_default_pg_num: 128
    osd_pool_default_pgp_num: 128

When the PG autoscaler is enabled (default in Quincy+), a new pool
starts with `pg_num = 1`. The configured `osd_pool_default_pgp_num: 128`
then conflicts with that effective `pg_num = 1`, triggering the ERANGE
check.

Upstream guidance is to *not* set `osd_pool_default_pgp_num` and let
the autoscaler manage both values:
- ceph-users ML: https://www.spinics.net/lists/ceph-users/msg76292.html
- Red Hat KB: https://access.redhat.com/solutions/7018605
- https://ceph.io/en/news/blog/2022/autoscaler_tuning/

Confirmed: removing \`osd_pool_default_pgp_num\` from the inventory
makes pool creation succeed without any further workaround.

## Fix

Remove \`osd_pool_default_pgp_num\` (and arguably
\`osd_pool_default_pg_num\`) from the example inventory.

dupremathieu:
## Summary
- Remove \`osd_pool_default_pg_num: 128\` and \`osd_pool_default_pgp_num: 128\` from the example SEAPATH cluster inventory.
- These overrides conflict with the PG autoscaler (enabled by default in Quincy+), which starts new pools at \`pg_num = 1\`. The mismatch makes pool creation fail with \`Error ERANGE: 'pgp_num' must be greater than 0 and lower or equal than 'pg_num', which in this case is 1\` on small clusters.
- Replaces #917, which worked around the symptom by creating pools with a fixed \`pg_num\` and re-enabling the autoscaler. The inventory-level fix removes the need for that workaround entirely.

Closes #922.

## Test plan
- [x] Deploy a 3-node SEAPATH cluster with small OSDs using the example inventory.
- [x] Verify the RBD pool is created without ERANGE errors.
- [x] Verify CephFS pools are created without EINVAL errors.
- [x] Verify \`ceph osd pool autoscale-status\` shows the autoscaler managing pg_num/pgp_num.

### DIFF
--- a/inventories/examples/seapath-cluster.yaml
+++ b/inventories/examples/seapath-cluster.yaml
@@ -109,8 +109,6 @@ mons:
       global:
         osd_pool_default_size: "{{ groups['hypervisors'] | length }}"
         osd_pool_default_min_size: 1
-        osd_pool_default_pg_num: 128
-        osd_pool_default_pgp_num: 128
         osd_crush_chooseleaf_type: 1
         mon_osd_min_down_reporters: 1
       mon:


dupremathieu:
Tested here: https://github.com/dupremathieu/seapath-virtual-sandbox/actions/runs/25369923098/job/74390486017

insatomcat (approved):

PR #924, opened 2026-05-07
cluster_setup_cephadm: add cephfs support for Yocto

Paullgk:
Add the needed modification to support Cephfs integration in SEAPATH Yocto. Note that:
- Required files `ceph-mds-ready@seapathcephfs.service` and `wait-for-mds.sh` are already installed in the image.
- The default Cephfs directory is located in the persistent directory.

**NOTE:** This PR is currently only compatible on SEAPATH Yocto Wrynose.

### DIFF
--- a/roles/deploy_cephfs/README.md
+++ b/roles/deploy_cephfs/README.md
@@ -11,6 +11,7 @@ You need to use cephadm and not ceph-ansible to use this feature
 | Variable                      | Required | Type   | Default                                    | Comments                                                                 |
 |-------------------------------|----------|--------|--------------------------------------------|--------------------------------------------------------------------------|
 | `deploy_cephfs_localdirtoupload` | No       | string | No default                                  | Local directory that will be uploaded to ceph_fs volume once created |
+| `deploy_cephfs_remotedir_map`    | No       | Map    | No default                                  | Destination directory of the CephFS mount point depending of the SEAPATH distribution |
 | `deploy_cephfs_remotedir`        | No       | string | `/mnt/cephfs/`                              | Destination directory on the remote hosts (CephFS mount point). Files from `deploy_cephfs_localdirtoupload` will be copied here. |
 
 ## Example Playbook
--- a/roles/deploy_cephfs/README.md
+++ b/roles/deploy_cephfs/README.md
@@ -8,11 +8,10 @@ You need to use cephadm and not ceph-ansible to use this feature
 
 ## Role Variables
 
-| Variable                      | Required | Type   | Default                                    | Comments                                                                 |
-|-------------------------------|----------|--------|--------------------------------------------|--------------------------------------------------------------------------|
-| `deploy_cephfs_localdirtoupload` | No       | string | No default                                  | Local directory that will be uploaded to ceph_fs volume once created |
-| `deploy_cephfs_remotedir_map`    | No       | Map    | No default                                  | Destination directory of the CephFS mount point depending of the SEAPATH distribution |
-| `deploy_cephfs_remotedir`        | No       | string | `/mnt/cephfs/`                              | Destination directory on the remote hosts (CephFS mount point). Files from `deploy_cephfs_localdirtoupload` will be copied here. |
+| Variable                      | Required | Type   | Default                                           | Comments                                                                 |
+|-------------------------------|----------|--------|---------------------------------------------------|--------------------------------------------------------------------------|
+| `deploy_cephfs_localdirtoupload` | No       | string | No default                                     | Local directory that will be uploaded to ceph_fs volume once created |
+| `deploy_cephfs_remotedir`        | No       | string | `/mnt/cephfs/` (/mnt/persistant/ceph on Yocto) | Destination directory on the remote hosts (CephFS mount point). Files from `deploy_cephfs_localdirtoupload` will be copied here. |
 
 ## Example Playbook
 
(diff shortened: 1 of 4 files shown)


roles/deploy_cephfs/tasks/main.yml
```diff
@@ -92,6 +92,7 @@
     mode: "0755"
     owner: root
     group: root
+  when: seapath_distro == "Debian"
```
eroussy: Remove this line.
If the file is already installed in the image at the correct location, ansible should just answer "ok"

roles/deploy_cephfs/tasks/main.yml
```diff
@@ -100,6 +101,7 @@
     mode: "0644"
     owner: root
     group: root
+  when: seapath_distro == "Debian"
```
eroussy: Remove this line.
If the file is already installed in the image at the correct location, ansible should just answer "ok"

roles/deploy_cephfs/tasks/main.yml
```diff
@@ -120,6 +122,11 @@
       }}
   run_once: true
 
+- name: Set deploy_cephfs_remotedir for yocto distro
```
eroussy: You should handle that in the defaults of the role IMO. Otherwise the user will not be able to overwrite it.
That beeing said, I don't know how to handle that the default is different between Yocto and Debian. May be use the seapath_distro variables mechanism ?

eroussy (changes requested):

roles/deploy_cephfs/tasks/main.yml
Paullgk: The files are not exactly the same (the path between Yocto and Debian are not identical). 
As the SHA256 sum is different, Ansible will keep trying to push the files on the target, leading to an error.

roles/deploy_cephfs/tasks/main.yml
Paullgk: The files are not exactly the same (the path between Yocto and Debian are not identical). 
As the SHA256 sum is different, Ansible will keep trying to push the files on the target, leading to an error.

roles/deploy_cephfs/tasks/main.yml
Paullgk: Done

roles/deploy_cephfs/defaults/main.yml
```diff
@@ -2,4 +2,8 @@
 # SPDX-License-Identifier: Apache-2.0
 
 ---
-deploy_cephfs_remotedir: "/mnt/cephfs/"
+deploy_cephfs_remotedir_map:
+  Debian: "/mnt/cephfs/"
+  Yocto: "/mnt/persistent/test/"
```
eroussy: This should be `/mnt/persistent/cephfs` ?

roles/deploy_cephfs/defaults/main.yml
```diff
@@ -2,4 +2,8 @@
...
+
+deploy_cephfs_remotedir: "{{ deploy_cephfs_remotedir_map[seapath_distro] | default('/mnt/cephfs/') }}"
```
eroussy: Add that to the documentation

eroussy (dismissed):

roles/deploy_cephfs/defaults/main.yml
Paullgk: Exact, fixed

dupremathieu (dismissed):
LGTM

roles/deploy_cephfs/README.md
```diff
@@ -11,6 +11,7 @@ You need to use cephadm and not ceph-ansible to use this feature
 | Variable                      | Required | Type   | Default                                    | Comments                                                                 |
 |-------------------------------|----------|--------|--------------------------------------------|--------------------------------------------------------------------------|
 | `deploy_cephfs_localdirtoupload` | No       | string | No default                                  | Local directory that will be uploaded to ceph_fs volume once created |
+| `deploy_cephfs_remotedir_map`    | No       | Map    | No default                                  | Destination directory of the CephFS mount point depending of the SEAPATH distribution |
```
eroussy: No, this is an internal variable. It should not appear
However, the line "deploy_cephfs_remotedir" should be changed to document the default : /mnt/cephfs for Debian and /mnt/persistent/cephfs for Yocto

eroussy (changes requested):

roles/deploy_cephfs/README.md
```diff
@@ -11,7 +11,6 @@ You need to use cephadm and not ceph-ansible to use this feature
 | Variable                      | Required | Type   | Default                                    | Comments                                                                 |
 |-------------------------------|----------|--------|--------------------------------------------|--------------------------------------------------------------------------|
 | `deploy_cephfs_localdirtoupload` | No       | string | No default                                  | Local directory that will be uploaded to ceph_fs volume once created |
-| `deploy_cephfs_remotedir_map`    | No       | Map    | No default                                  | Destination directory of the CephFS mount point depending of the SEAPATH distribution |
 | `deploy_cephfs_remotedir`        | No       | string | `/mnt/cephfs/`                              | Destination directory on the remote hosts (CephFS mount point). Files from `deploy_cephfs_localdirtoupload` will be copied here. |
```
eroussy: Change the default to "`/mnt/persistend/cephfs` (Yocto) or `/mnt/cephfs/` (Debian)"

eroussy (changes requested):

eroussy (approved):

insatomcat (approved):

PR #926, opened 2026-05-11
roles/cephadm: add missing default variable

dupremathieu:
cephadm_installpackage default variable was missing.


dupremathieu:
Merged in #927

PR #927, opened 2026-05-11
fix: various fixes

dupremathieu:
## Summary

- Fix string-to-boolean conditional error in `configure_libvirt_rdb_secret` by adding `| bool` filter
- Fix `delegate_to` crash in `configure_ha` when `valid_machine` group doesn't exist
- Fix errors with Ansible 2.20

### configure_libvirt_rdb_secret
The Jinja2 template in the `set_fact` task produces a string `"true"` instead of a boolean. Adding `| bool` to the `when` conditions fixes the "Conditional result (True) was derived from value of type 'str'" error.

### configure_ha
The `delegate_to: "{{ groups['valid_machine'][0] }}"` expression fails with "object of type 'dict' has no attribute 'valid_machine'" when no hosts were added to that group. Using `groups.get()` with a fallback avoids the crash. The `when` condition still prevents the task from actually running.

### DIFF
--- a/roles/configure_ha/tasks/main.yml
+++ b/roles/configure_ha/tasks/main.yml
@@ -75,7 +75,7 @@
     - groups['valid_machine'] is defined
     - groups['unconfigured_machine_group'] is defined
   run_once: true
-  delegate_to: "{{ groups['valid_machine'][0] }}"
+  delegate_to: "{{ (groups.get('valid_machine') or [inventory_hostname]) | first }}"
   block:
     - name: Fetch corosync key
       ansible.builtin.fetch:
--- a/library/cluster_vm.py
+++ b/library/cluster_vm.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/python3
 
 # Copyright (C) 2021, RTE (http://www.rte-france.com)
 # SPDX-License-Identifier: Apache-2.0
--- a/roles/configure_libvirt_rdb_secret/tasks/main.yml
+++ b/roles/configure_libvirt_rdb_secret/tasks/main.yml
@@ -18,7 +18,7 @@
     configure_libvirt_rdb_secret_create_secret_pool: "{% if hostvars[item]['configure_libvirt_rdb_secret_secret_defined'].stdout == '' %}true{% else %}{{ configure_libvirt_rdb_secret_create_secret_pool | default(false) }}{% endif %}"
   loop: "{{ groups['hypervisors'] | intersect(groups['cluster_machines']) }}"
 
-- when: configure_libvirt_rdb_secret_create_secret_pool
+- when: configure_libvirt_rdb_secret_create_secret_pool | bool
   block:
     - name: Check if the secret is already defined
       ansible.builtin.set_fact:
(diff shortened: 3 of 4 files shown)


roles/configure_ha/tasks/main.yml
```diff
@@ -75,7 +75,7 @@
     - groups['valid_machine'] is defined
     - groups['unconfigured_machine_group'] is defined
   run_once: true
-  delegate_to: "{{ groups['valid_machine'][0] }}"
+  delegate_to: "{{ (groups.get('valid_machine') or [inventory_hostname]) | first }}"
```
insatomcat, line 78: if the group valid_machine does not exist, then the `groups['valid_machine'] is defined` makes the code ignore this block... so why is it a problem?
this block is run on the CI all the time with no existing cluster (so with no valid_machine group) and it does not crash.

dupremathieu: Delegate_to is evaluating before when: https://github.com/ansible/ansible/issues/15746

Without this hook I have the following error:
```
TASK [configure_ha : Fetch corosync key] ****************************************************************************************************************************************************************************
Tuesday 19 May 2026  10:04:27 +0200 (0:00:01.001)       0:04:07.354 ***********                                                                                                                                      
[ERROR]: Task failed: object of type 'dict' has no attribute 'valid_machine'                                                                                                                                         
                                                                                                                                                                                                                     
Task failed.                                                                                                                                                                                                         
Origin: /home/mdupre/Documents/seapath/seapath-virtual-sandbox/ansible/roles/configure_ha/tasks/main.yml:80:7                                                                                                        
                                                                                                                                                                                                                     
78   delegate_to: "{{ groups['valid_machine'][0] }}"                                                                                                                                                                 
79   block:                                                                                                                                                                                                          
80     - name: Fetch corosync key                                                                                                                                                                                    
         ^ column 7                                                                                                                                                                                                  
                                                                                                                                                                                                                     
<<< caused by >>>                                                                                                                                                                                                    
                                                                                                                                                                                                                     
object of type 'dict' has no attribute 'valid_machine'                                                                                                                                                               
Origin: /home/mdupre/Documents/seapath/seapath-virtual-sandbox/ansible/roles/configure_ha/tasks/main.yml:78:16                                                                                                       
                                                                                                                                                                                                                     
76     - groups['unconfigured_machine_group'] is defined                                                                                                                                                             
77   run_once: true                                                                                                                                                                                                  
78   delegate_to: "{{ groups['valid_machine'][0] }}"                                                                                                                                                                 
                  ^ column 16                                                                                                                                                                                        
                                                                                                                                                                                                                     
fatal: [node1 -> {{ groups['valid_machine'][0] }}]: FAILED! => {"changed": false, "msg": "Task failed: object of type 'dict' has no attribute 'valid_machine'"}                      
```

insatomcat: ok but do you have any idea why it never failed before on the CI?
anything specific on your setup?

dupremathieu: It seems like it's due to the Ansible version.
It's passed on version 2.16 but not on 2.20.

Since we are currently on version 2.16 I can remove this hook or keep it the comment as you want.

insatomcat: keep it, no worries.

insatomcat (approved):

PR #928, opened 2026-05-13
roles: split cephadm role into cephadm_install and cephadm

dupremathieu:
The cephadm role previously handled both binary installation and cluster provisioning. This blocked ceph_prepare_installation which needs the cephadm to be setup (it call cephadm ceph-volume) but must run before the cluster is set up.

cephadm_install now runs first on cluster_machines, providing the binary before ceph_prepare_installation prepares OSD disks. The cephadm role then handles only cluster bootstrapping and provisioning.


roles/cephadm_install/tasks/main.yml
```diff
@@ -0,0 +1,118 @@
+# Copyright (C) 2026 Savoir-faire Linux, Inc.
```
insatomcat, line 2: This content was split out from roles/cephadm/tasks/main.yml, moving code doesn't transfer authorship, so the original copyright should be kept.

insatomcat (approved):

PR #930, opened 2026-05-20
Support remote VM system disks in deploy_vms_cluster

dahoat-sprecher:
Hello to All,

during deployment of VMs, we wanted to decouple Ansible repo location from VM disk image location, resulting in this Pull Request. 

**Intent**
Let operators provision VMs from a system disk image hosted on a remote URL, fetched directly by the target node using the `ansible.builtin.get_url` module, instead of staging the qcow2 from the Ansible control machine.

**Purpose / Why**
- Removes the need to host every VM image on the control node and push it over SSH or deploy it independently before creating the VM.
- Adds integrity verification so a remote source can be trusted: a checksum is mandatory for remote disks and optional (but validated) for local ones.

**What changes (roles/deploy_vms_cluster)**
- New per-VM inventory variables:
  - `vm_remote_disk` — URL of the image, downloaded on the target via `get_url`. Mutually exclusive with `vm_disk`.
  - `vm_disk_checksum` — `dict {<algo>: <hex>}`, required for `vm_remote_disk`, optional for `vm_disk`. Accepts sha256/sha384/sha512, the strongest provided algorithm is used.
- tasks/main.yml:
  - Validates that exactly one of vm_disk / vm_remote_disk is set.
  - Rejects unknown checksum keys and malformed checksum dicts.
  - Downloads via get_url with checksum: for remote disks, keeps the existing copy path for local disks.
  - After staging, re-verifies the file via stat + assert so a mismatch fails the play with a clear message and regardless of the image origin. May be changed to only do this on the `vm_disk` path as `get_url` already checked this after fetching from a remote location. 
- README updated with the two new variables and the mutual-exclusion / checksum rules.

All the best, 
Daniel


dahoat-sprecher:
Pull request superseded by #1004

PR #934, opened 2026-05-21
debian hardening: Disable iptables for bridging when docker is enabled

### LINKED ISSUE #909: Docker br_netfilter silently blocks KVM VM bridge traffic on SEAPATH nodes
## Summary

On SEAPATH nodes running Docker alongside KVM virtual machines, the `br_netfilter` kernel module (loaded by Docker) causes all bridged traffic — including KVM VM traffic — to pass through iptables. libvirt's default iptables FORWARD policy is DROP. The result is that a KVM VM can be running, its network interface visible on the bridge, its MAC address learned via ARP — but all IP traffic is silently dropped.

This was discovered during deployment of an ABB vSSC600 virtual protection relay on a SEAPATH Debian standalone node (kernel 6.1.0-41-rt-amd64, libvirt 9.0.0, QEMU 7.2.19).

## Relationship to #379

Closed issue #379 ("Unable to access VM when using a linux bridge on hypervisor") describes the same symptom — a VM connected to a Linux bridge is unreachable despite correct configuration. That issue was closed with a workaround of using an OVS bridge instead. The root cause (`br_netfilter` + iptables FORWARD DROP) was not identified at the time. This issue documents the actual mechanism and proposes a platform-level fix.

## How to Reproduce

1. Deploy a SEAPATH Debian standalone node with Docker installed
2. Create a Linux bridge (e.g. `mgmt-bridge`) using systemd-networkd
3. Deploy a KVM VM with a bridge interface on `mgmt-bridge`
4. From an external machine on the same L2 segment, attempt to ping the VM's IP address
5. Observe: ARP resolves (VM MAC is visible) but ping and all higher-layer traffic fails

## How to Confirm

```bash
lsmod | grep br_netfilter
cat /proc/sys/net/bridge/bridge-nf-call-iptables
# If br_netfilter loaded and value = 1 — affected

sudo iptables -L FORWARD -n -v
# Will show DROP policy and incrementing dropped packet count
```

## Fix

```bash
# Disable immediately
sudo sysctl -w net.bridge.bridge-nf-call-iptables=0
sudo sysctl -w net.bridge.bridge-nf-call-ip6tables=0
sudo sysctl -w net.bridge.bridge-nf-call-arptables=0

# Persist across reboots
echo 'net.bridge.bridge-nf-call-iptables = 0' | sudo tee -a /etc/sysctl.conf
echo 'net.bridge.bridge-nf-call-ip6tables = 0' | sudo tee -a /etc/sysctl.conf
echo 'net.bridge.bridge-nf-call-arptables = 0' | sudo tee -a /etc/sysctl.conf
```

This fix only affects bridged traffic. Docker continues to function normally — Docker uses NAT/masquerade for container networking, not bridging.

## Proposed Platform Change

Add these three sysctl settings to the SEAPATH hardening playbook or node provisioning process as a standard configuration item on nodes that run Docker alongside KVM VMs. This would silently resolve the conflict for all future deployments without manual intervention.

## Environment

- SEAPATH Debian standalone, kernel 6.1.0-41-rt-amd64
- KVM / libvirt 9.0.0 / QEMU 7.2.19
- Docker installed (loads br_netfilter)
- VM: ABB vSSC600 SW v1.5.1 with Linux bridge interface

Reported by @ni8towl

vliaskov:
Fixes: https://github.com/seapath/ansible/issues/909

@ni8towl it would be good to validate the change on your platform, thanks for offering.

I assume the changed sysctl settings should also be checked in `roles/debian_tests/cukinia-tests/cukinia/common_security_tests.d/sysctl.conf`  . But the settings change is conditional on docker service being enabled.

### DIFF
--- a/roles/debian_hardening/tasks/main.yml
+++ b/roles/debian_hardening/tasks/main.yml
@@ -82,6 +82,55 @@
   notify: Update sysfs values using sysctl
   when: revert
 
+- name: "Check if docker service exists on the target"
+  ansible.builtin.service_facts:
+- name: "Ensure br_netfilter loads before systemd-sysctl"
+  ansible.builtin.copy:
+    dest: /etc/modules-load.d/50-bridge-nf-call.conf
+    content: "br_netfilter\n"
+    owner: root
+    group: root
+    mode: '0644'
+  when:
+    - not revert
+    - ansible_facts['services']['docker.service']['status'] == 'enabled'
+- name: "Disable iptables for bridge" # noqa: risky-file-permissions
+  ansible.builtin.lineinfile:
+    dest: /etc/sysctl.d/50-bridge-nf-call-iptables.conf
+    regexp: "{{ item.regexp }}"
+    line: "{{ item.line }}"
+    create: true
+    state: present
+  loop:
+    - { regexp: '^net.bridge.bridge-nf-call-iptables=0$', line: 'net.bridge.bridge-nf-call-iptables=0' }
+    - { regexp: '^net.bridge.bridge-nf-call-ip6tables=0$', line: 'net.bridge.bridge-nf-call-ip6tables=0' }
+    - { regexp: '^net.bridge.bridge-nf-call-arptables=0$', line: 'net.bridge.bridge-nf-call-arptables=0' }
+  notify: Update sysfs values using sysctl
+  when:
+    - not revert
+    - ansible_facts['services']['docker.service']['status'] == 'enabled'
+- name: "Revert disabling iptables for bridge"
+  ansible.builtin.file:
+    dest: /etc/sysctl.d/50-bridge-nf-call-iptables.conf
+    state: absent
+  when:
+    - revert
+    - ansible_facts['services']['docker.service']['status'] == 'enabled'
+- name: "Restore kernel defaults at revert time"
+  ansible.posix.sysctl:
+    name: "{{ item }}"
+    value: '1'
+    state: present
+    sysctl_set: true
+    reload: false
+  loop:
+    - net.bridge.bridge-nf-call-iptables
+    - net.bridge.bridge-nf-call-ip6tables
+    - net.bridge.bridge-nf-call-arptables
+  when:
+    - revert
+    - ansible_facts['services']['docker.service']['status'] == 'enabled'
+
 - name: "Install hardened sysfs rules" # noqa: risky-file-permissions
   ansible.builtin.copy:
     src: sysctl/90-sysctl-hardening.conf


ni8towl:
Hi @vliaskov — thanks for picking up #909. Validated the fix on a Debian 12 + Docker test environment and it resolved the bug at value level. The test also surfaced one structural issue that's worth addressing before merge: the fix doesn't survive a reboot under the standard Debian package's load ordering. Detail and proposed remediation below, with boot-timeline evidence.

### Value-level fix — confirmed

Writing the three sysctls into `/etc/sysctl.d/50-bridge-nf-call-iptables.conf` and running `sysctl --system` resolved the silent-drop behaviour immediately. Same topology, same `iptables -P FORWARD DROP`, ping outcome inverted purely by the sysctl change:

| Test run | Sysctls | 5-ping result | Main-ns FORWARD counter delta |
|---|---|---|---|
| Pre-fix | 1 / 1 / 1 | 100% loss, exit 1 | **5 packets / 420 bytes** |
| Post-fix (same session) | 0 / 0 / 0 | 100% delivery, exit 0 | **0 / 0** |

### Persistence defect — load ordering

After a reboot, with the fix file still on disk and unchanged, the runtime sysctls returned to `1` and the ping again dropped 5/5 with the same 5 / 420 counter signature. The boot journal shows why:

```log
22:55:43.447989  systemd-sysctl[267]: Couldn't write '0' to 'net/bridge/bridge-nf-call-iptables',
                                       ignoring: No such file or directory
22:55:43.448009  systemd-sysctl[267]: Couldn't write '0' to 'net/bridge/bridge-nf-call-ip6tables',
                                       ignoring: No such file or directory
22:55:43.448019  systemd-sysctl[267]: Couldn't write '0' to 'net/bridge/bridge-nf-call-arptables',
                                       ignoring: No such file or directory
22:55:43.592775  systemd[1]: Finished systemd-sysctl.service - Apply Kernel Variables.
                              [status=0/SUCCESS]
22:55:49.921373  systemd[1]: Starting docker.service                            (+6.3 s)
22:55:50.220837  kernel: bridge: filtering via arp/ip/ip6tables is no longer
                          available by default. Update your scripts to load
                          br_netfilter if you need this.
22:55:50.709322  systemd[1]: Started docker.service
```

`systemd-sysctl.service` runs under `sysinit.target` (early). `docker.service` runs under `multi-user.target` (later). On a stock Debian 12 + `docker.io` install, nothing causes `br_netfilter` to load before `systemd-sysctl.service` runs — the `docker.io` package ships no `/etc/modules-load.d/` entry, and `dpkg -L docker.io | grep -iE 'modules-load|modprobe'` confirms no postinst module-loading hooks. So `/proc/sys/net/bridge/*` does not exist when `sysctl --system` reads `/etc/sysctl.d/`, the three writes fail with `ENOENT`, systemd-sysctl logs "ignoring" and exits cleanly with `status=0/SUCCESS`. Docker starts ~6 s later, loads `br_netfilter` — which initialises its tunables to the kernel defaults of `1` — and nothing re-asserts the values from the sysctl file. The fix is silently lost on every boot.

The same signature appears on an existing production SEAPATH host that has an independently-developed manual `/etc/sysctl.conf` fix for this bug:

```log
May 23 18:40:17 seapath systemd-sysctl[402]: Couldn't write '0' to 'net/bridge/bridge-nf-call-iptables', ignoring: No such file or directory
May 23 18:40:17 seapath systemd-sysctl[402]: Couldn't write '0' to 'net/bridge/bridge-nf-call-ip6tables', ignoring: No such file or directory
May 23 18:40:17 seapath systemd-sysctl[402]: Couldn't write '0' to 'net/bridge/bridge-nf-call-arptables', ignoring: No such file or directory
```

Identical fingerprint, different file locations (`/etc/sysctl.conf` vs `/etc/sysctl.d/50-...conf`), same root cause. PR #934 as currently drafted has the same boot-time gap — it works for the Ansible-apply session (because the handler runs `sysctl --system` after `docker.service` has already loaded `br_netfilter`), but the next boot loses the fix again.

### Proposed remediation — empirically validated

Adding one file at `/etc/modules-load.d/50-bridge-nf-call.conf` containing the single line `br_netfilter` puts module loading under `systemd-modules-load.service`, which runs at sysinit.target strictly before `systemd-sysctl.service`. The keys exist when sysctl reads them; the writes succeed:

```log
23:00:37.462983  systemd-modules-load[259]: Inserted module 'br_netfilter'
23:00:37.568104  systemd[1]: Starting systemd-modules-load.service
23:00:37.568305  kernel: bridge: filtering via arp/ip/ip6tables is no longer available...
23:00:37.568372  systemd[1]: Finished systemd-modules-load.service
23:00:37.568383  systemd[1]: Starting systemd-sysctl.service
23:00:37.568549  systemd[1]: Finished systemd-sysctl.service        [no error output → clean]
23:00:43.150273  systemd[1]: Starting docker.service                (+5.6 s)
23:00:43.825861  systemd[1]: Started docker.service
```

Two reboots, same VM, same `50-bridge-nf-call-iptables.conf`, same Docker install. The only difference between the failing boot above and the succeeding boot here is that one file in `/etc/modules-load.d/`. Post-boot sysctls correctly at `0 / 0 / 0`, ping succeeds, FORWARD counter delta zero.

Two equivalently-correct implementations to add to `roles/debian_hardening/tasks/main.yml`:

```yaml
# Option A — mirrors the existing sysctl file pattern in this PR
- name: "Ensure br_netfilter loads before systemd-sysctl"
  ansible.builtin.copy:
    dest: /etc/modules-load.d/50-bridge-nf-call.conf
    content: "br_netfilter\n"
    owner: root
    group: root
    mode: '0644'
  when:
    - not revert
    - ansible_facts['services']['docker.service']['status'] == 'enabled'
```

```yaml
# Option B — more idiomatic; single task handles "load now" and "load at boot"
- name: "Load br_netfilter now and on every boot"
  community.general.modprobe:
    name: br_netfilter
    state: present
    persistent: present
  when:
    - not revert
    - ansible_facts['services']['docker.service']['status'] == 'enabled'
```

The filename in `/etc/modules-load.d/` is cosmetic; the `50-` prefix was chosen to match the sysctl file's prefix for visual symmetry but anything alphabetically anywhere works. Either is equally correct; the choice is yours.

### Two refinements to the revert path

**Revert leaves an empty file behind.** The current revert task uses three loop iterations of `ansible.builtin.lineinfile` with `state: absent`. Each iteration strips one of the three lines from `/etc/sysctl.d/50-bridge-nf-call-iptables.conf`; after the third, the file exists at 0 bytes. `sysctl --system` ignores it, but operators auditing `/etc/sysctl.d/` will see a dangling artefact. Replacing the three-task loop with one `ansible.builtin.file: state=absent` removes the whole file in one primitive.

**Runtime sysctls do not revert when the file is removed.** This was unexpected. `sysctl --system` only writes values found in config files; it does not reset values whose declarations have been removed. After the revert task runs:

```log
$ sudo sysctl --system
* Applying /etc/sysctl.d/50-bridge-nf-call-iptables.conf ...    (empty/absent file)
... (other files apply normally)
                                                                  (no bridge-nf line in output)
$ sysctl net.bridge.bridge-nf-call-iptables
net.bridge.bridge-nf-call-iptables = 0                            (still 0)
```

The runtime values stay where they were until either a reboot or an explicit write. An operator running the revert play and expecting `sysctl` to immediately show `1` will be surprised. Adding three explicit `ansible.posix.sysctl` writes to the revert task closes this gap:

```yaml
- name: "Restore kernel defaults at revert time"
  ansible.posix.sysctl:
    name: "{{ item }}"
    value: '1'
    state: present
    sysctl_set: true
    reload: false
  loop:
    - net.bridge.bridge-nf-call-iptables
    - net.bridge.bridge-nf-call-ip6tables
    - net.bridge.bridge-nf-call-arptables
  when:
    - revert
    - ansible_facts['services']['docker.service']['status'] == 'enabled'
```

### Smaller observations

**Don't add `modprobe -r br_netfilter` to the revert.** On a host with no other bridge users, `modprobe -r br_netfilter` cascade-unloads the `bridge` module (because `br_netfilter` is its only refcount holder), which destroys all bridge interfaces in the namespace — including `docker0`. Docker's daemon stays running but its bridge is ripped out; container networking is broken until `systemctl restart docker`. The current PR has no such call; this is a note to anyone tempted to add one.

**The bug is wider than "libvirt + Docker collision".** Installing `docker.io` on a clean Debian 12 — with no libvirt involvement, before any container starts — flips `iptables -P FORWARD` from ACCEPT to DROP as part of Docker's own isolation chains. The SEAPATH-specific manifestation comes from libvirt VM bridges colliding with Docker, but the precondition (`FORWARD=DROP` + `br_netfilter` + sysctls=1) is present on *any* Debian + Docker host. Worth keeping in mind when scoping the fix's eligibility and Changelog entry.

**Cukinia regression.** The check in `roles/debian_tests/cukinia-tests/cukinia/common_security_tests.d/sysctl.conf` (already flagged in the PR description) will fail after merge on any host that meets the `docker.service` condition, since it expects the kernel defaults. Needs a matching update conditioned on the same Docker presence check.

### Methodology

Test environment: Debian 12 (`debian-12-generic-amd64.qcow2`, kernel `6.1.0-48-amd64`), `docker.io 20.10.24`, in a fresh KVM/libvirt VM provisioned for this validation on a SEAPATH home lab. Reproduction used a port-to-port bridge topology — `br-test` with no IP carrying two veth pairs into two network namespaces (`ns_host` 10.99.99.1, `ns_vm` 10.99.99.2) — because that's the minimal setup that exercises the bridge-forwarding path where `br_netfilter` intercepts into iptables FORWARD. `iptables -P FORWARD DROP` was set explicitly in main ns. Serial console was logged to file for boot-stage visibility. The same defect-class fingerprint observed in the test VM also appears on an existing production SEAPATH host carrying an independently-developed manual `/etc/sysctl.conf` version of this same fix — both fail in the same boot-time `ENOENT` window.

Scope caveat: this validates the behaviour of the fix on a stock Debian 12 + Docker host. The full Ansible role mechanics (idempotency, role-ordering, interaction with the rest of debian_hardening) are out of scope for this exercise and remain CI's territory.

vliaskov:
Thanks for the extensive testing and the good catches @ni8towl  

I updated the branch of the PR with the following changes:

- reboot issue regarding sysctl value setting and module loading: I added a new task with option A, though this might be up to maintainer preference.
- revert path: I added a task for reverting the sysctl values to 1 as you suggested.
- revert path: I changed the task to simply deleting the whole file as you suggested.

dupremathieu:
Your PR fix a Docker issue which is not supported by SEAPATH (we use Podman).

You can use `upload_extra_files` in your inventory to upload your files.

PR #943, opened 2026-06-11
Virtual cluster CI trigger

Paullgk:
Hello,
This PR adds the trigger of the virtual-cluster CI, which allows to test the PR in a SEAPATH Yocto virtual cluster.

### DIFF
--- a/.github/workflows/ci-virtual-cluster.yml
+++ b/.github/workflows/ci-virtual-cluster.yml
@@ -13,6 +13,3 @@ on:
 jobs:
   virtual-cluster:
     uses: seapath/virtual-cluster/.github/workflows/ansible-setup.yml@main
-    with:
-      ansible_repo: https://github.com/seapath/ansible.git
-      ansible_ref: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || 'main' }}
--- a/.github/workflows/ci-virtual-cluster.yml
+++ b/.github/workflows/ci-virtual-cluster.yml
@@ -8,6 +8,7 @@ name: CI Virtual cluster
 
 on:
   workflow_dispatch:
+  workflow_call:
 
 jobs:
   virtual-cluster:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,3 +36,7 @@ jobs:
   ci-centos:
     needs: ci-molecule
     uses: ./.github/workflows/ci-centos.yml
+
+  ci-virtual-cluster:
+    needs: ci-molecule
+    uses: ./.github/workflows/ci-virtual-cluster.yml


.github/workflows/ci-virtual-cluster.yml
```diff
@@ -14,5 +14,5 @@ jobs:
   virtual-cluster:
     uses: seapath/virtual-cluster/.github/workflows/ansible-setup.yml@main
     with:
-      ansible_repo: https://github.com/seapath/ansible.git
+      ansible_repo: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.repo.clone_url || 'https://github.com/seapath/ansible.git' }}
```
eroussy: Do we really want to launch the virtual cluster with different ansible repository ?
For me, the only use case is to launch it from seapath ansible repo, with a PR eventually coming from a fork
This is already covered by a classic actions/checkout

Paullgk: In this case, how can we test the modification provided by the PR?

.github/workflows/ci.yml
```diff
@@ -36,3 +36,7 @@ jobs:
   ci-centos:
     needs: ci-molecule
     uses: ./.github/workflows/ci-centos.yml
+
+  ci-virtual-cluster:
+    needs: ci-molecule
```
dupremathieu, line 41: Do we really need ci-molecule?

.github/workflows/ci-virtual-cluster.yml
eroussy: It is precisely the job of action/checkout to download the code from the PR when a workflow is triggered `on pull_request` 
`ci.yml` set up some hidden configuration variables like GITHUB_REF and other kind. It is then used by the other workflow.
It is already done like that for `ci-qa` and `ci-molecule` if you want to have a look

dupremathieu: Maybe we can do the opposite.
The action lives in this repo and this repo clone virtual-cluster.

.github/workflows/ci-virtual-cluster.yml
```diff
@@ -8,10 +8,11 @@ name: CI Virtual cluster
...
 on:
   workflow_dispatch:
+  workflow_call:
 
 jobs:
   virtual-cluster:
     uses: seapath/virtual-cluster/.github/workflows/ansible-setup.yml@main
     with:
-      ansible_repo: https://github.com/seapath/ansible.git
+      ansible_repo: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.repo.clone_url || 'https://github.com/seapath/ansible.git' }}
```
dupremathieu: We want one job per flavors.

dupremathieu: We also need to do all the CI stuff except latencies tests.

The workflow should be something like that.

- install prerequisist
- checkout ansible
- checkout virtual-cluser
- download image
- configure tvars
- in virtual-cluser `make init apply start ansible-ping`
- next run regular CI files

.github/workflows/ci-virtual-cluster.yml
Paullgk: I modified the behavior of the virtual-cluster workflow in order to use the Ansible repo cloned and checkout itself, and not the one cloned in the virtual-cluster workflow.

.github/workflows/ci.yml
Paullgk: Every other CI Actions needs it, I replicated the same behavior

.github/workflows/ci-virtual-cluster.yml
Paullgk: It has been reworked see the latest commits

Paullgk: I agree, but might be done also in the CI script refactor.
For now I suggest to merge a first iteration to test virtual cluster setup on each PR?

eroussy:
Commit message 3 needs a bit more explanation, otherwise LGTM

.github/workflows/ci-virtual-cluster.yml
eroussy: Yes, specifically, the report can be done afterwards

eroussy (approved):

eroussy (approved):

PR #947, opened 2026-06-15
Make hardening configuration roles generic

emontmas:
Currently, SEAPATH Debian has a dedicated hardening role, configuring a number of SEAPATH security requirements.

This PR adapt the role to make it generic. This will ease integration of the role with any new SEAPATH flavor. This is necessary to use the role with SEAPATH SLES for example.

* Rename roles and variables for `configure_hardening` and `configure_hardening_physical_machine` roles.
* Split tasks in sub-files for improved readability.
* Introduce new role variables that are dependent to the underlying distribution.
* Rename hardening playbooks to be coherent with other playbooks name.
* Configure SEAPATH SLES for the hardening playbook.

Note that this PR takes a distro-agnostic approach in the role configuration. Instead of having per-distro var files inside the role directory (e.g. `vars/Debian.yml`, `vars/CentOS.yml`), roles expect the calling playbook to set the role variables beforehand. In this PR, this is done inside the prerequisites playbooks, where role variables are given when calling the role.
* Separating distro-variables from roles better fit the reusable purpose of a role.
* Per-distro variables values are less scattered across the repository, which improves readability and maintainability.
* Adding a new SEAPATH distro is easier, as only the new playbook calling the roles need to be written.

Therefore, hardening variables are put into dedicated per-distro files at the repo-level (`vars` directory) and loaded by the playbooks that needs them.

Related to https://github.com/seapath/ci/pull/130


roles/configure_hardening/README.md
```diff
@@ -17,6 +17,7 @@ No requirement.
 | ip_addr         | Yes      | String | IP address for administration   |
 | cluster_ip_addr | No       | String | Cluster IP address              |
 | configure_hardening_ssh_service | Yes | String | Name of SSH systemd service |
+| configure_hardening_login_defs_file | No | The login.defs file to configure. Defaults to /etc/login.defs |
```
eroussy: You should add a default column in this table and provide the default value of login_defs_file

roles/configure_hardening/README.md
```diff
@@ -18,6 +18,7 @@ No requirement.
 | cluster_ip_addr | No       | String | Cluster IP address              |
 | configure_hardening_ssh_service | Yes | String | Name of SSH systemd service |
 | configure_hardening_login_defs_file | No | The login.defs file to configure. Defaults to /etc/login.defs |
+| configure_hardening_etc_securetty_group | No | The group of the /etc/securetty file. Defaults to root |
```
eroussy: Provide default value in README

eroussy (changes requested):
I am not a fan of having only one hardening playbook.
As in previous PR, you kept multiple prerequisites playbook (one per distribution), it make more sense to have multiple hardening playbook (one per distribution)
It will be simpler to maintain and will avoid intensive use of multiple seapath_distro stuff
Some distro may not even use hardening (Yocto, may be SUSE if we only rely on open scap one day)
What do you think ?

eroussy:
Also, one really cool feature to add would be a verify step for required variables to be set, and fail with meaningful message if it is not the case.
You have an example in the [configure_nic_irq_affinity](https://github.com/seapath/ansible/tree/main/roles/configure_nic_irq_affinity/tasks)
Do you think you have time to do that ?

roles/configure_hardening/README.md
emontmas: Done, also added value type

roles/configure_hardening/README.md
emontmas: Done, also added value type

emontmas:
@eroussy 

> I am not a fan of having only one hardening playbook. As in previous PR, you kept multiple prerequisites playbook (one per distribution), it make more sense to have multiple hardening playbook (one per distribution) It will be simpler to maintain and will avoid intensive use of multiple seapath_distro stuff Some distro may not even use hardening (Yocto, may be SUSE if we only rely on open scap one day) What do you think ?

I agree, but it brings the question more globally for the other roles, that:
* Currently use the `vars/<distro>.yml` files
* Are part of other playbooks and called by `seapath_setup_main.yml`.
For example, the `snmp` role used in `seapath_setup_snmp.yaml`.
And for that I don't think it is scalable to create a variant of every possible playbook for all distro. We can't have five `seapath_setup_snmp.yaml` playbooks, five `cluster_setup_ha.yaml` playbooks, five `seapath_setup_main.yaml`, ... This would be unmaintainable.

My recommendation would be some kind of middle ground. Keep a global per-distro file in `vars/`, but which contains every role input variable set in it. Then this file is dynamically given to each role with input variables through a `vars_files` parameter.

That way at least all distro-specific variables are gathered at the same place, and roles in themselves stay generic and non-distro specific.

The eventual alternative I see would still using common playbook, but each time a role needs parameters, it is called with multiple `import_role` tasks, one for each distro, with a `when` statement based of `seapath_distro` and distro specific variables as input. I personally don't like it as it would make the playbooks very verbose.

I would suggest opening an issue as this PR focus on the integration of the hardening roles to SEAPATH SLES.

This is my opinion on the matter of course.
Ultimately, this needs you, the maintainers, to discuss how to harmonise the roles and playbook for the best so that it stays maintainable as new SEAPATH flavors are integrated.

emontmas:
@eroussy 

> Also, one really cool feature to add would be a verify step for required variables to be set, and fail with meaningful message if it is not the case. You have an example in the [configure_nic_irq_affinity](https://github.com/seapath/ansible/tree/main/roles/configure_nic_irq_affinity/tasks) Do you think you have time to do that ?

I won't have time to do this now sorry.
But this is also a matter of global harmonisation of the repository's roles. I suggest opening a PR and letting you, the maintainers, first decide on how to do it, related to my above remarks about roles that still have per-distro variable files.

Also, note that since ansible-core 2.11, Ansible has a built-in feature for role parameter validation: https://docs.ansible.com/projects/ansible/latest/playbook_guide/playbooks_reuse_roles.html#role-argument-validation
Maybe this would be better to validate role input variables?

eroussy:
> @eroussy
> 
> > I am not a fan of having only one hardening playbook. As in previous PR, you kept multiple prerequisites playbook (one per distribution), it make more sense to have multiple hardening playbook (one per distribution) It will be simpler to maintain and will avoid intensive use of multiple seapath_distro stuff Some distro may not even use hardening (Yocto, may be SUSE if we only rely on open scap one day) What do you think ?
> 
> I agree, but it brings the question more globally for the other roles, that:
> 
>     * Currently use the `vars/<distro>.yml` files
> 
>     * Are part of other playbooks and called by `seapath_setup_main.yml`.
>       For example, the `snmp` role used in `seapath_setup_snmp.yaml`.
>       And for that I don't think it is scalable to create a variant of every possible playbook for all distro. We can't have five `seapath_setup_snmp.yaml` playbooks, five `cluster_setup_ha.yaml` playbooks, five `seapath_setup_main.yaml`, ... This would be unmaintainable.
> 
> 
> My recommendation would be some kind of middle ground. Keep a global per-distro file in `vars/`, but which contains every role input variable set in it. Then this file is dynamically given to each role with input variables through a `vars_files` parameter.
> 
> That way at least all distro-specific variables are gathered at the same place, and roles in themselves stay generic and non-distro specific.
> 
> The eventual alternative I see would still using common playbook, but each time a role needs parameters, it is called with multiple `import_role` tasks, one for each distro, with a `when` statement based of `seapath_distro` and distro specific variables as input. I personally don't like it as it would make the playbooks very verbose.
> 
> I would suggest opening an issue as this PR focus on the integration of the hardening roles to SEAPATH SLES.
> 
> This is my opinion on the matter of course. Ultimately, this needs you, the maintainers, to discuss how to harmonise the roles and playbook for the best so that it stays maintainable as new SEAPATH flavors are integrated.

Agreed with all you said.
But I think we should align prerequisites and hardening playbook. If we have multiple prerequisites playbook, we should also have multiple hardening playbooks.
The rest of the playbooks should be in common

@dupremathieu What do you think ?

emontmas:
> > Also, one really cool feature to add would be a verify step for required variables to be set, and fail with meaningful message if it is not the case. You have an example in the [configure_nic_irq_affinity](https://github.com/seapath/ansible/tree/main/roles/configure_nic_irq_affinity/tasks) Do you think you have time to do that ?
> 
> I won't have time to do this now sorry. But this is also a matter of global harmonisation of the repository's roles. I suggest opening a PR and letting you, the maintainers, first decide on how to do it, related to my above remarks about roles that still have per-distro variable files.
> 
> Also, note that since ansible-core 2.11, Ansible has a built-in feature for role parameter validation: https://docs.ansible.com/projects/ansible/latest/playbook_guide/playbooks_reuse_roles.html#role-argument-validation Maybe this would be better to validate role input variables?

@eroussy Issue opened here: https://github.com/seapath/ansible/issues/950

eroussy (approved):

emontmas:
> Agreed with all you said. But I think we should align prerequisites and hardening playbook. If we have multiple prerequisites playbook, we should also have multiple hardening playbooks. The rest of the playbooks should be in common

@eroussy @dupremathieu Issue opened on the questions of what to do with distro-specific role variables, and the split of hardening playbooks: https://github.com/seapath/ansible/issues/952

PR #948, opened 2026-06-15
Add SEAPATH SLES cluster support

TanguyRaufflet:
This PR adds the missing pieces to deploy the cluster configurations on SEAPATH SLES images.
It includes:
- The ability to customize the Ceph user and group IDs for each distribution
- The Pacemaker RA
- The path to the crm executable
- A fix for the team0_x template service


roles/cephadm_install/defaults/main.yml
```diff
@@ -11,3 +11,8 @@ cephadm_install_repo: "{{ cephadm_installrepo | default(false) }}"
 cephadm_install_common: "{{ cephadm_installcommon | default(false) }}"
 cephadm_install_package: "{{ cephadm_installpackage | default(false) }}"
 cephadm_install_pullimages: "{{ cephadm_pullimages | default(false) }}"
+cephadm_install_cephadm_uid: "{{ omit }}"
```
eroussy: Idiomatic way to declare a required variable is to put a null default:
`cephadm_install_cephadm_uid: null`

playbooks/cluster_setup_cephadm.yaml
```diff
@@ -4,12 +4,21 @@
 # This playbook configures and sets up Ceph with cephadm.
 
 ---
```
eroussy, line 6: If you use seapath_distro variable, you must set it with detect_seapath_distro role

eroussy (changes requested):

roles/cephadm_install/defaults/main.yml
TanguyRaufflet: I chose  "omit" because the uid/gid properties are optional.
But you're right, it's probably better to force all the distribution to set those variables.

playbooks/cluster_setup_cephadm.yaml
TanguyRaufflet: Done

eroussy (dismissed):

eroussy (dismissed):

eroussy (approved):

PR #955, opened 2026-06-19
Add cukinia tests support for SEAPATH SLES

emontmas:
Up until now, the cukinia tests in SEAPATH Ansible were dedicated to SEAPATH Debian.
As SEAPATH SLES joins the list of SEAPATH flavors, and in a effort to share cukinia tests as much as possible to avoid code duplication, this PR adapts the existing cukinia tests and roles for SEAPATH SLES.

* The `debian_tests` role is renamed `deploy_cukinia_tests`. It will be a common role to SEAPATH Debian and SEAPATH SLES.
  * In the future, this opens the possibility to:
    * Merge the SEAPATH Yocto cukinia tests in this role, which are very similar. That way, only one cukinia code base will need to be maintained, improving coherence of the common SEAPATH tests across all flavors.
    * Supporting the cukinia tests for newer SEAPATH flavors will be easier and quicker to do.
* The test deployment role is improved to automatically template and copy any jinja test file, removing the need to maintain a list of `.j2` files in the task file. This is needed, as making a common role means tests files needs the flexibility of jinja for every distro-specific configurations.
* Fix group and user tests so that the related tests:
  * Have meaningful names, representing what they do.
  * Are not dependent from other test results, which is not clearly conveyed in a test report.
* Adapt cukinia tests for SEAPATH SLES:
  * Users and groups tests.
  * Services tests.
  * Setuid / setgid test.
  * File paths.
  * Package test.

Note that for now a few tests have been identified as failed in SEAPATH SLES and unfixable from SEAPATH. These tests have therefore been explicitly set as "skipped" for SEAPATH SLES, using cukinia's keyword `when`. This was done to avoid making them fail any cukinia test run without completely silencing them.
* `SEAPATH-00050 -- Linux kernel 'hardening' : LEGACY_VSYSCALL_NONE is enabled`
  * This kernel configuration isn't enabled in the `kernel-rt` package used in SEAPATH SLES 16.0. This is a build time kernel configuration and cannot be fixed by the SEAPATH project.
* `SEAPATH-00084 -- syslog-ng is configured to send log on network` --> checks for existence of `/sys/kernel/realtime`
  * As specified in the related commit, this file is specific to the Linux kernel RT patch and hasn't been merged in the mainline despite official RT support in mainline kernel since version 6.12. Nevertheless, this file is merely an alternative to checking for `PREEMPT_RT` in the kernel configuration. So it does specifically check if the kernel is RT **and uses the RT patch**. As SEAPATH SLES 16.0 cleary uses a mainline RT kernel, skip this test.

Other minor fixes more general to SEAPATH ansible that come with the cukinia tests adaptation for SLES:
* Fixed permissions of deployed cukinia test files.
* In `ansible-requirements.yaml` specify a version condition for the `community.general` collection that officially support ansible-core 2.16, which the version currently in used for SEAPATH Ansible.
* In the example inventories, change the grub password example to use 65536 rounds as required by the ANSSI BP28 related cukinia test.
* In the `configure_hypervisor` role, add a task to enable `libvirtd.service`. This is the default setting for Debian, but not for SLES and potentially for other distros.
* Fix file permission in the `configure_hardening` role, according to hardening cukinia tests.
* Add sshd config customization needed for SEAPATH SLES in the `configure_hardening` role.
* Make sure the auditd configuration allows the same logging format as expected by cukinia tests.


roles/deploy_cukinia_tests/cukinia-tests/cukinia/common_security_tests.d/packages_tests.d/sles.conf.j2
```diff
@@ -0,0 +1,200 @@
...
+            if ! echo "$REQUIRED_PACKAGES" | grep -Eq "(^|\s)$r(\s|$)"; then
+                REQUIRED_PACKAGES="$REQUIRED_PACKAGES $r"
+            fi
+        done
+
+        return 0
+    fi
+
+    # Detect packages installed with weak requirements (e.g. recommends, suggest, ...)
+    requires="$(zypper --no-remote -q se -i --requires-pkg $1 | tail -n+4 | cut -d'|' -f2)"
```
TanguyRaufflet, line 168: Why no using zypper packages --userinstalled or `zypper search --installed-only --type package | grep '^i+'` ?

roles/deploy_cukinia_tests/cukinia-tests/cukinia/common_security_tests.d/packages_tests.d/sles.conf.j2
```diff
@@ -0,0 +1,200 @@
...
+    bzip2 \
+    ca-certificates \
+    ca-certificates \
+    ca-certificates-mozilla \
+    ceph-common \
+    cephadm \
+    chrony \
+    cockpit \
+    cockpit-machines \
+    cockpit-packagekit \
```
TanguyRaufflet, line 23: The Cockpit packages should be listed in the "ESSENTIAL_PACKAGES" ?

roles/deploy_cukinia_tests/cukinia-tests/cukinia/common_tests.d/preempt-rt.conf
```diff
@@ -1,7 +1,15 @@
...
 test_id "SEAPATH-00075" as "kernel is PREEMPT RT" cukinia_cmd grep "PREEMPT_RT" /proc/version
-test_id "SEAPATH-00076" as "kernel is realtime" cukinia_test $(cat /sys/kernel/realtime) -eq 1
+
+if [ -f /sys/kernel/realtime ]; then
+    SYS_KERNEL_REALTIME_CONTENT="$(cat /sys/kernel/realtime)"
+else
+    SYS_KERNEL_REALTIME_CONTENT=0
+fi
+
+when "[ -f /sys/kernel/realtime ]" test_id "SEAPATH-00076" as "kernel is using the RT patch" cukinia_test $SYS_KERNEL_REALTIME_CONTENT -eq 1
```
TanguyRaufflet, line 15: If the file `/sys/kernel/realtime` is not a good indicator to identify if a RT kernel is used, maybe we can use something else instead ?
(Ex: "-rt" in `uname -r` OR "PREEMPT_RT" in `uname -a` ?)

roles/deploy_cukinia_tests/cukinia-tests/cukinia/common_security_tests.d/packages_tests.d/sles.conf.j2
emontmas: Because your commands won't do the same thing.

Here what I basically need to do is a dependency tree, considering any installed packages as leaves (packages that are required by other).
So in one instance of the recursive function, what I try to now is "Which packages requires package `$1`?", which is searching for reverse requirements. This is basically what `zypper --no-remote -q se -i --requires-pkg $1` does. And then either one of the "package that needs `$1` to be installed" is part of the "essential packages list", or I need to recursively check these packages, in case their are multiple levels of requirements (E.g. A that needs B that need C, with A being an essential package).

`zypper packages --userinstalled` or `zypper search --installed-only --type package | grep '^i+'` will give me what are the package which were manually installed, which is something completely different.

roles/deploy_cukinia_tests/cukinia-tests/cukinia/common_security_tests.d/packages_tests.d/sles.conf.j2
emontmas: I used the same naming scheme as for the Debian package test, but maybe "ESSENTIAL" isn't the best. This list is more "What are the packages that are expected to be found on the system", so that the test checks that all installed packages have a reason to be there.

roles/deploy_cukinia_tests/cukinia-tests/cukinia/common_tests.d/preempt-rt.conf
emontmas: The above test in this file already checks for `PREEMPT_RT` in the distro string, using `/proc/version` instead of `uname -a`

So the "Check that RT kernel" test was in fact duplicated with these two tests. And as said in the commit message, this second test is more "Is the kernel RT and using the RT patch"

roles/deploy_cukinia_tests/cukinia-tests/cukinia/cluster_tests.d/vm_manager_libvirt.conf.j2
```diff
@@ -13,8 +13,10 @@ clean
 
 cukinia_log "$(_colorize yellow "--- Test vm_manager module libvirt part ---")"
 
+{% if inventory_hostname in groups.get('hypervisors', []) %}
```
eroussy: actually, libvirt should not be installed on observers, so this complete file should be skipped for observers

eroussy (changes requested):

roles/deploy_cukinia_tests/cukinia-tests/cukinia/cluster_tests.d/vm_manager_libvirt.conf.j2
emontmas: Ack, I will change that
Note however that currently libvirt is installed on observers on Debian and Yocto, and it should be fixed in later PRs. I think an issue was opened regarding this.

roles/deploy_cukinia_tests/cukinia-tests/cukinia/common_security_tests.d/packages_tests.d/sles.conf.j2
emontmas: Done, `ESSENTIAL_PACKAGES` renamed in `EXPECTED_PACKAGES`

roles/deploy_cukinia_tests/cukinia-tests/cukinia/cluster_tests.d/vm_manager_libvirt.conf.j2
emontmas: Done, test file is run only on hypervisors

TanguyRaufflet (approved):

eroussy (approved):

PR #957, opened 2026-06-22
Move cyclictest task to dedicated role

emontmas:
Currently, the `ci_yocto/run_tests` role contained tasks to run a cyclictest test on SEAPATH machines.
This isn't practical for reuse in other context, such as with other SEAPATH flavors or for development.

Therefore this PR:
* Extract the cyclictest tasks to a dedicated role.
* Add a role variable to configure the test duration.


roles/cyclictest/defaults/main.yml
```diff
@@ -0,0 +1,5 @@
+# Copyright (C) 2026 Savoir-faire Linux, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+---
+cukinia_test_prefix: ".."
```
eroussy: Rename to `cyclictest_test_prefix`

eroussy (changes requested):

roles/cyclictest/defaults/main.yml
emontmas: Done

dupremathieu (approved):

eroussy (approved):

PR #958, opened 2026-06-23
Update hardening configurations and add the OpenSCAP support for SLES

TanguyRaufflet:
This pull request adds support for OpenSCAP remediation, which will be used on SEAPATH SLES16 to complement the default hardening currently implemented in SEAPATH.

After applying the OpenSCAP remediation, it appeared that some hardening configurations present in Ansible did not match the ANSSI BP28 requirements. These configurations have been updated in this PR to avoid the hardening being applied twice.

### DIFF
--- a/playbooks/seapath_setup_hardening.yaml
+++ b/playbooks/seapath_setup_hardening.yaml
@@ -37,6 +37,19 @@
   roles:
     - configure_hardening_physical_machine
 
+- name: Run OpenSCAP remediation
+  become: true
+  hosts:
+    - cluster_machines
+    - standalone_machine
+  vars:
+    revert: false
+  pre_tasks:
+    - name: Load distribution-specific hardening variables
+      ansible.builtin.include_vars: ../vars/{{ seapath_distro }}_hardening.yml
+  roles:
+    - configure_hardening_oscap
+
 - name: Reboot to apply hardening
   become: true
   hosts:
--- a/roles/configure_hardening_oscap/tasks/main.yml
+++ b/roles/configure_hardening_oscap/tasks/main.yml
@@ -1,0 +1,21 @@
+# Copyright (C) 2026 Savoir-faire Linux, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+---
+- name: Run OpenSCAP ANSSI BP28 High remediation
+  ansible.builtin.command:
+    cmd: >
+      oscap xccdf eval
+      --remediate
+      --fetch-remote-resources
+      --results /tmp/oscap-results.xml
+      --report /tmp/oscap-report.html
+      --profile {{ configure_hardening_oscap_profile }}
+      {{ configure_hardening_oscap_ds_file }}
+  register: configure_hardening_oscap_remediation
+  failed_when: configure_hardening_oscap_remediation.rc not in [0, 2]
+  changed_when: configure_hardening_oscap_remediation.rc == 2
+  when:
+    - not revert
+    - configure_hardening_oscap_ds_file is defined
+    - configure_hardening_oscap_profile is defined
--- a/roles/configure_hardening_oscap/tasks/main.yml
+++ b/roles/configure_hardening_oscap/tasks/main.yml
@@ -11,6 +11,9 @@
       --results /tmp/oscap-results.xml
       --report /tmp/oscap-report.html
       --profile {{ configure_hardening_oscap_profile }}
+      {% if configure_hardening_oscap_tailoring_file is defined %}
+      --tailoring-file {{ configure_hardening_oscap_tailoring_file }}
+      {% endif %}
       {{ configure_hardening_oscap_ds_file }}
   register: configure_hardening_oscap_remediation
   failed_when: configure_hardening_oscap_remediation.rc not in [0, 2]
--- a/roles/configure_hardening/files/sudoers/security
+++ b/roles/configure_hardening/files/sudoers/security
@@ -1,4 +1,4 @@
-Defaults noexec ,requiretty ,use_pty ,umask=0027 ,ignore_dot ,env_reset
+Defaults noexec ,requiretty ,use_pty ,umask=0077 ,ignore_dot ,env_reset
 
 # These tags are required by the ANSSI-BP-028-R39
 # This file is integrated as 00-security on the target in order to be read
--- a/roles/configure_hardening/templates/01_password.j2
+++ b/roles/configure_hardening/templates/01_password.j2
@@ -1,4 +1,4 @@
 #!/usr/bin/sh
 
-echo superusers="root"
+echo set superusers="root"
 echo password_pbkdf2 root {{ grub_password }}
--- a/roles/configure_hardening/vars/main.yml
+++ b/roles/configure_hardening/vars/main.yml
@@ -16,6 +16,7 @@ configure_hardening_kernel_params:
   - mce=0
   - rng_core.default_quality=500
   - lsm=apparmor,lockdown,capability,landlock,yama,bpf,integrity
+  - kvm.nx_huge_pages=auto
 
 configure_hardening_hardened_services:
   - syslog-ng
--- a/roles/configure_hardening/vars/main.yml
+++ b/roles/configure_hardening/vars/main.yml
@@ -7,7 +7,7 @@ configure_hardening_kernel_params:
   - init_on_free=1
   - slab_nomerge
   - pti=on
-  - slub_debug=ZF
+  - slub_debug=ZFP
   - randomize_kstack_offset=on
   - slab_common.usercopy_fallback=N
   - iommu=pt
--- a/roles/configure_hardening/vars/main.yml
+++ b/roles/configure_hardening/vars/main.yml
@@ -11,6 +11,7 @@ configure_hardening_kernel_params:
   - randomize_kstack_offset=on
   - slab_common.usercopy_fallback=N
   - iommu=pt
+  - intel_iommu=on
   - security=yama
   - mce=0
   - rng_core.default_quality=500
--- a/roles/configure_hardening_oscap/README.md
+++ b/roles/configure_hardening_oscap/README.md
@@ -1,0 +1,24 @@
+# configure_hardening_oscap Role
+
+This role runs an OpenSCAP XCCDF remediation after the main hardening has been applied. It must be run **after** `configure_hardening` (and `configure_hardening_physical_machine` where applicable) so that OpenSCAP can remediate any remaining non-compliant settings. `openscap` and `scap-security-guide` must be installed on the target host.
+
+## Role Variables
+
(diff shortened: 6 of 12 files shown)


roles/configure_hardening/defaults/main.yml
```diff
@@ -5,3 +5,6 @@
 configure_hardening_login_defs_file: /etc/login.defs
 configure_hardening_etc_securetty_group: root
 configure_hardening_sshd_has_lastlog: true
+
+configure_hardening_oscap_profile: xccdf_org.ssgproject.content_profile_anssi_bp28_high
+configure_hardening_oscap_ds_file: /usr/share/xml/scap/ssg/content/ssg-sle16-ds.xml
```
emontmas: This profile cannot be a default one as it is specific to sles 16.0 according to the name.

If OpenSCAP remediation is still specific to SLES 16.0 for now, you should make it only happen with SLES 16.0.

For example, instead of setting a default here, set the variables specific to SLES in the existing file `vars/SLES_hardening.yml`
And then in the task file, you can conditionally execute the openscap tasks only if the related variables are set

I don't know what exactly is the `configure_hardening_oscap_profile` variable, I let you decide if this is something specific to SLES or if it would be the same for other distros and then make the correct change if needed

roles/configure_hardening/tasks/main.yml
```diff
@@ -38,3 +38,5 @@
   ansible.builtin.import_tasks: syslog.yml
 - name: Configure audit hardening
   ansible.builtin.import_tasks: audit.yml
+- name: Run OpenSCAP remediation
+  ansible.builtin.import_tasks: oscap.yml
```
emontmas: See https://github.com/seapath/ansible/pull/958/changes/2ba93cb0f544f87973b7e787a17ed30692ba8991#r3458046680 for conditional execution. Don't forget that you can't use `when` with `import_tasks` and should instead use `include_tasks`

roles/configure_hardening/tasks/oscap.yml
```diff
@@ -0,0 +1,20 @@
...
+      --results /tmp/oscap-results.xml
+      --report /tmp/oscap-report.html
+      --profile {{ configure_hardening_oscap_profile }}
+      {{ configure_hardening_oscap_ds_file }}
+  register: oscap_remediation
+  failed_when: oscap_remediation.rc not in [0, 2]
+  changed_when: oscap_remediation.rc == 2
+  when:
+    - not revert
+    - ansible_distribution == "SLES"
```
emontmas: Avoid when possible distro conditions in roles directly.
Instead rely on input variables, and let the calling context set them depending on the distro. See https://github.com/seapath/ansible/pull/958/changes/2ba93cb0f544f87973b7e787a17ed30692ba8991#r3458046680

Also, to allow possible easier integration in the future with SEAPATH Yocto, don't use `ansible_distribution` but `seapath_distro`

roles/configure_hardening/tasks/oscap.yml
```diff
@@ -0,0 +1,20 @@
...
+  ansible.builtin.command:
+    cmd: >
+      oscap xccdf eval
+      --remediate
+      --fetch-remote-resources
+      --results /tmp/oscap-results.xml
+      --report /tmp/oscap-report.html
+      --profile {{ configure_hardening_oscap_profile }}
+      {{ configure_hardening_oscap_ds_file }}
+  register: oscap_remediation
```
emontmas: Variables should have `<role>_` prefix, see lint

roles/configure_hardening/defaults/main.yml
```diff
@@ -6,5 +6,6 @@ configure_hardening_login_defs_file: /etc/login.defs
 configure_hardening_etc_securetty_group: root
 configure_hardening_sshd_has_lastlog: true
 
-configure_hardening_oscap_profile: xccdf_org.ssgproject.content_profile_anssi_bp28_high
+configure_hardening_oscap_profile: xccdf_org.ssgproject.content_profile_anssi_bp28_high_customized
```
emontmas: This change is not related to the commit `configure_hardening: oscap.yml: add support for tailoring file` and should be done in the previous one when adding `configure_hardening_oscap_profile`

roles/configure_hardening/defaults/main.yml
```diff
@@ -6,5 +6,6 @@ configure_hardening_login_defs_file: /etc/login.defs
...
 configure_hardening_oscap_ds_file: /usr/share/xml/scap/ssg/content/ssg-sle16-ds.xml
+configure_hardening_oscap_tailoring_file: /usr/share/xml/scap/ssg/content/ssg-sle16-ds-tailoring.xml
```
emontmas: Same remark as https://github.com/seapath/ansible/pull/958/changes/da33a883ad76afc73cff3e08989ab1a911e58776#r3458046680

emontmas (changes requested):

roles/configure_hardening/defaults/main.yml
TanguyRaufflet: This change really related to this commit. When using a tailoring file, the name of the openScap must be change to reflect the new profile created [1], which is an extension of the default one (i.e. content_profile_anssi_bp28_high)

[1]: https://github.com/seapath/kiwi-descriptions/blob/bb57e1e2d8719f8021abe7659303300a9253d60e/kiwi-descriptions/x86_64/sle/seapath-sles16.0/root/usr/share/xml/scap/ssg/content/ssg-sle16-ds-tailoring.xml#L5

emontmas: Then maybe specify this in the commit message?

roles/configure_hardening/defaults/main.yml
TanguyRaufflet: Done.

roles/configure_hardening/tasks/oscap.yml
TanguyRaufflet: I replaced this condition, to not rely on the distro used.

roles/configure_hardening/tasks/oscap.yml
TanguyRaufflet: Done.

roles/configure_hardening/defaults/main.yml
TanguyRaufflet: Commit message updated.

roles/configure_hardening/defaults/main.yml
TanguyRaufflet: See the discussion above.

roles/configure_hardening_oscap/tasks/main.yml
emontmas, line 1: If these tasks are not meant to be called from the `main.yml` task file, or isn't an entry-point for other tasks in the `configure_hardening` role, it feels like it is a separate role

I think it would be best to put this in a dedicated role, like `configure_hardening_oscap` or `oscap_hardening`

playbooks/seapath_setup_hardening.yaml
```diff
@@ -37,6 +37,22 @@
...
+- name: Run OpenSCAP remediation
+  become: true
+  hosts:
+    - cluster_machines
+    - standalone_machine
+  vars:
+    revert: false
+  pre_tasks:
+    - name: Load distribution-specific hardening variables
+      ansible.builtin.include_vars: ../vars/{{ seapath_distro }}_hardening.yml
```
emontmas, line 49: Play syntax already has a `vars_files` keyword for that, see https://docs.ansible.com/projects/ansible/latest/reference_appendices/playbooks_keywords.html#play

```suggestion
  vars_files:
    - ../vars/{{ seapath_distro }}_hardening.yml
```

Note that you may have to test to be sure that variable expansion is possible at the play level (I have a doubt)

roles/configure_hardening_oscap/tasks/main.yml
emontmas: I know you had to make sure oscap stuff happen after all the hardening, but if it is that deconnected from the role it should probably be a separate role in itself, so that it will be more straitghforward to use

roles/configure_hardening_oscap/tasks/main.yml
```diff
@@ -0,0 +1,20 @@
...
+      --results /tmp/oscap-results.xml
+      --report /tmp/oscap-report.html
+      --profile {{ configure_hardening_oscap_profile }}
+      {{ configure_hardening_oscap_ds_file }}
+  register: configure_hardening_oscap_remediation
+  failed_when: configure_hardening_oscap_remediation.rc not in [0, 2]
+  changed_when: configure_hardening_oscap_remediation.rc == 2
+  when:
+    - not revert
+    - configure_hardening_oscap_ds_file is defined
```
emontmas, line 23: Don't you need to also check that `configure_hardening_oscap_profile` exists ?

roles/configure_hardening/tasks/oscap.yml
```diff
@@ -11,6 +11,9 @@
       --results /tmp/oscap-results.xml
       --report /tmp/oscap-report.html
       --profile {{ configure_hardening_oscap_profile }}
+      {% if configure_hardening_oscap_tailoring_file | length > 0 %}
```
emontmas: Won't this fail if `configure_hardening_oscap_tailoring_file` isn't set? In that case you need to check existence with `configure_hardening_oscap_tailoring_file is defined`

emontmas (changes requested):

playbooks/seapath_setup_hardening.yaml
TanguyRaufflet: `vars_files` is not used in this repo.
I don't think we should change how variables are included in this specific case. Also, including variables in the pre_tasks with ansible.builtin.include_vars is more robust.

roles/configure_hardening_oscap/tasks/main.yml
TanguyRaufflet: Because I considered both must be provided, if not it should return an error.
But it's also valid to add configure_hardening_oscap_profile.

--> Suggestion applied

roles/configure_hardening/tasks/oscap.yml
TanguyRaufflet: Done

roles/configure_hardening_oscap/tasks/main.yml
TanguyRaufflet: Yes you're right, I've created a dedicated role.

playbooks/seapath_setup_hardening.yaml
emontmas: I see that I myself use the `include_vars` syntax in previous plays a few weeks earlier.
So OK for me to align with what was already done.
I don't know why I didn't used the `vars_files` syntax, but it may be worth it using it in the future

emontmas (approved):

eroussy (approved):

PR #959, opened 2026-06-23
add deploy_prometheus_exporters role

insatomcat:
Provide a self-contained role to deploy Prometheus exporters as Podman quadlet units, with per-host exporter selection via inventory variables.


playbooks/seapath_setup_prometheus_exporters.yaml
```diff
@@ -0,0 +1,35 @@
...
+- name: Deploy prometheus exporters on observers
+  hosts: observers
+  become: true
+  vars:
+    deploy_prometheus_exporters_exporters: "{{ deploy_prometheus_exporters_profile_minimal }}"
+  roles:
+    - deploy_prometheus_exporters
+
+- name: Deploy prometheus exporters on VMs
+  hosts: VMs
```
dupremathieu: I think we should have a dedicated playbook for VMs because not all VMs will support it.

roles/deploy_prometheus_exporters/templates/insatomcat-exporter.container.j2
```diff
@@ -0,0 +1,19 @@
+[Unit]
+Description=Prometheus insatomcat Exporter (vhost CPU monitoring)
```
dupremathieu, line 2: Remove this insatomcat and change the file name.

dupremathieu (changes requested):

roles/deploy_prometheus_exporters/templates/insatomcat-exporter.container.j2
insatomcat: I think it's a useful exporter, I did it myself, it allows monitoring the vhost threads cpu consumption, something the libvirt api does not expose.
https://hub.docker.com/r/insatomcat/insatomcat-exporter

playbooks/seapath_setup_prometheus_exporters.yaml
insatomcat: ok

roles/deploy_prometheus_exporters/templates/insatomcat-exporter.container.j2
dupremathieu: Ha ok. I've thought you have forgotten to remove instomcat.

insatomcat: if you don't any other remark, can you approve plz ?

dupremathieu: Sorry I haven't completely finished the reviewing.

insatomcat:
Hello @dahoat-sprecher, @jaxx0r,
You offered to contribute to Action 3.1 of the roadmap, which includes the Prometheus integration.
This PR proposes what I believe should be the first step: deploying some exporters.
Please don't hesitate to share your feedback and/or approve it if you think it looks good.
Thanks!

dahoat-sprecher:
Hello @insatomcat , 
thanks for the invitation and notification. I am looking forward to having a look as soon as possible. Unfortunately, I won't be able until after my vacation. :sweat_smile:

dahoat-sprecher:
Hello @insatomcat , 

I now had the chance to test this branch and it worked perfectly for me. 
However, one thing caught my eye: Did you already plan on how to integrate TLS and authentication?

Best regards, 
Daniel

insatomcat:
Hi @dahoat-sprecher,

Thanks a lot for taking the time to test it, glad it worked out of the box for you.

TLS and authentication: good question, and to be honest it is something we still need to think through properly rather than something already settled. I left it out of this first step on purpose. There are clear leads though, and the good news is that the groundwork is already there.

Every exporter deployed by this role can serve its metrics over TLS and verify a client certificate. For the upstream ones this comes from `prometheus/exporter-toolkit`, which gives us a single common configuration format through `--web.config.file`:

```yaml
tls_server_config:
  cert_file: /etc/prometheus/exporters/tls/server.crt
  key_file:  /etc/prometheus/exporters/tls/server.key
  client_auth_type: RequireAndVerifyClientCert
  client_ca_file: /etc/prometheus/exporters/tls/ca.crt
  min_version: TLS13
```

So the likely direction is one such file per host, mounted read-only into the containers, with the role exposing a single set of variables and taking care of the small configuration differences between exporters. No reverse proxy, no sidecar, no extra failure point.

I currently lean towards mutual TLS rather than basic auth for the authentication part. It keeps any shared secret out of the inventory, revocation goes through the CA, and it avoids having to generate and vault bcrypt hashes. It is also uniformly available across the exporters we deploy. That said it is a preference and not a decision, so I would genuinely like your view on it.

For certificate distribution, this repository already has a pattern worth following: the `syslog_ng_client` and `vmmgrapi` roles take certificate and key paths as variables and copy them into place. Reusing it would keep things consistent instead of inventing a mechanism specific to monitoring.

One part will not fit that model at all: the Ceph side, since the mgr prometheus module is not an exporter-toolkit binary. There the path looks more like `mgr/cephadm/secure_monitoring_stack`, which deserves its own discussion.

I would rather keep all of this for a follow-up PR. Working iteratively keeps each PR small enough to be reviewed and converged on quickly, and in my experience that gets us to the end result faster than one large change that stalls in review.

So, any other remark on the current state, or do you think we can merge this one and move the security work to the next step? Feedback very welcome on the design above, and you are of course more than welcome to take part in it.

dahoat-sprecher:
Yes, I fully agree on the usage of mTLS, I am only afraid of the complexity of handling the certificates and will look into the existing approaches. 

Two items might deserve a second look which I try to comment directly.

roles/deploy_prometheus_exporters/templates/node-exporter.container.j2
```diff
@@ -0,0 +1,20 @@
...
+Image={{ deploy_prometheus_exporters_node_exporter_image }}
+PodmanArgs=--pid=host
+PodmanArgs=--network=host
+PodmanArgs=--hostname=%H
+Volume=/:/host:ro,rslave
+Volume=/var/run/dbus:/var/run/dbus:ro
+Exec=--path.rootfs=/host --collector.textfile.directory=/host/var/lib/prometheus/node_exporter --web.listen-address={{ deploy_prometheus_exporters_listen_address }}:9100 --collector.systemd
+
+[Service]
+Restart=unless-stopped
```
dahoat-sprecher: As far as I know, this value should be `always`. The default value might be `Restart=no` on invalid values.

roles/deploy_prometheus_exporters/tasks/main.yml
```diff
@@ -0,0 +1,94 @@
...
+
+    - name: Flush handlers before starting exporter services
+      ansible.builtin.meta: flush_handlers
+      when: deploy_prometheus_exporters_manage_services | bool
+
+    - name: Enable and start podman API socket
+      ansible.builtin.systemd:
+        name: podman.socket
+        enabled: true
+        state: started
```
dahoat-sprecher, line 55: If I understand this correctly, the service is only started if not already running. Therefore, changes on the template are not picked up until an explicit restart of the service or the machine (same for "Enable and start prometheus exporter services on #[Line 61](https://github.com/seapath/ansible/pull/959/changes#diff-25340fb6b45784e731f14888634aa81b674866cee85f0ba1e2fc4e4653f90d8cR61). 

Claude claimed there is a workaround to combine looped items with handlers:
```yaml
# tasks/main.yml
- name: Deploy prometheus exporter container units
  ansible.builtin.template:
    src: "{{ item }}.container.j2"
    dest: "/etc/containers/systemd/{{ item }}.container"
    ...
  loop: "{{ deploy_prometheus_exporters_exporters }}"
  register: deploy_prometheus_exporters_unit_files
  notify:
    - Reload systemd after exporter container unit changes
    - Restart changed exporter services

# handlers/main.yml
- name: Reload systemd after exporter container unit changes
  ansible.builtin.systemd:
    daemon_reload: true

- name: Restart changed exporter services
  ansible.builtin.systemd:
    name: "{{ item }}"
    state: restarted
  loop: "{{ deploy_prometheus_exporters_unit_files.results | selectattr('changed') | map(attribute='item') | list }}"
  when: deploy_prometheus_exporters_manage_services | bool
```

roles/deploy_prometheus_exporters/templates/node-exporter.container.j2
insatomcat: Good catch, thanks. `unless-stopped` is a Podman/Docker restart policy, not a systemd one: systemd only accepts `no`/`on-success`/`on-failure`/`on-abnormal`/`on-watchdog`/`on-abort`/`always`, and Quadlet copies the `[Service]` section verbatim into the generated unit. I confirmed it on one of our lab machines that already runs these exporters:

```
/etc/containers/systemd/node-exporter.container:  Restart=unless-stopped
systemctl show -p Restart node-exporter:          Restart=no
systemctl show -p Restart podman-exporter:        Restart=no
```

So the effective policy really was `Restart=no`, exactly as you suspected.

Fixed to `Restart=always` + `RestartSec=5`. While at it I aligned the other templates: `ha_cluster_exporter.container.j2` and `podman-exporter.container.j2` had no `[Service]` section at all, so they were on `Restart=no` too. All five exporters now share the same policy, and the molecule verify asserts it so a regression gets caught.

After redeploying the fixed role on that machine, `systemctl kill --signal=SIGKILL node-exporter` gives `NRestarts=1` and the exporter is back up and scrapable a few seconds later.

roles/deploy_prometheus_exporters/tasks/main.yml
insatomcat: You are right, and this is a real bug: `state: started` is a no-op on an already running service, so a template change was written to disk and the unit regenerated, but the container kept running with the old configuration until a manual restart or a reboot. The role was not convergent.

I applied essentially your snippet: the template task is now registered and notifies both handlers, and a new `Restart changed exporter services` handler restarts only the units whose file actually changed.

Two details worth noting on top of it:

- handlers run in declaration order, not in notification order, so the daemon-reload stays declared first in `handlers/main.yml` (there is a comment saying so now);
- I moved the `podman.socket` task before the `flush_handlers`, and dropped the `when` on `meta: flush_handlers`. The `manage_services` condition now lives on the restart handler itself, which is where it belongs, and a daemon-reload on its own is harmless.

I also had a look at the `enabled: true` sitting next to it, since Quadlet units live in `/run/systemd/generator/` and systemd normally refuses to enable a generated unit. It turns out to be a silent no-op rather than a failure: `systemctl is-enabled` returns `generated` with rc 0, and Ansible's systemd module only calls `systemctl enable` for `enabled-runtime`/`indirect`/`alias`, so it never runs. Since it can never do anything on a Quadlet unit, and reading it suggests the role handles boot activation when the `[Install]` section does, I dropped it from the exporter task. The one on `podman.socket` stays, that one is a real unit.

Tested on a lab machine that was already running the exporters. First run: both `.container` files change, the reload and restart handlers fire, and the effective policy goes from `Restart=no` to `Restart=always` with the services restarted, no reboot involved. Second run: `changed=0` and no handler, so the role stays idempotent.

dahoat-sprecher: The changes look good to me, ~~however, I cannot resolve the comment.~~

dahoat-sprecher (approved):

insatomcat:
hi @jaxx0r, any feedback on your side?

dahoat-sprecher:
Another thing regarding SELinux: I have never worked with it, i am just forwarding the message here: 
The `node-exporter` is mounting `/` and `libvirt-exporter` the libvirt socket while both are unprivileged and `SecurityLabelDisable` is not set. 
Therefore, it could be that SELinux blocks on CentOS and Oracle Linux, but I cannot test this as I have neither set up at the moment. 

In case you also want to move this to a future PR, I am completely fine with it.

insatomcat:
No need to defer it, we have taken it into account.

It does not bite on our images: the SEAPATH kickstart for Oracle Linux 9 boots with `selinux=0`. But the role was inconsistent with itself anyway. insatomcat-exporter and podman-exporter already set `SecurityLabelDisable=true`, and ha_cluster_exporter runs `--privileged`, which implies it. Only node-exporter and libvirt-exporter were missing it, so both set it now, just pushed. A relabel is not an option here: `:z` on `Volume=/:/host` would recursively relabel the whole host filesystem.

insatomcat:
@npochet-elia FYI, since you asked to be informed on point 3.1 of the roadmap

insatomcat:
All stakeholders identified in roadmap section 3.1 have been contacted and informed. (Note: I initially included @jaxxor in this group, but he clarified that he will not be participating.)

We're awaiting your approval, @dupremathieu and @eroussy.

dahoat-sprecher:
I just noticed a faulty RAM module on one of my cluster machines resulting in many correctable errors and occasional crashes (currently 1/day). 

Do we already export metrics helping in such a situation? A quick search for "memory" in the exported metrics revealed nothing of interest. 
I guess, we should use this opportunity as long as its not completely dead :sweat_smile:.

insatomcat:
> Do we already export metrics helping in such a situation? A quick search for "memory" in the exported metrics revealed nothing of interest.

Can you search for "edac" in the node exporter metrics?
`curl $ip:9100/metrics | grep edac` (Error Detection And Correction)

dahoat-sprecher:
Perfect, thank you

image

dahoat-sprecher:
Did you also consider adding a LVM exporter, for example https://github.com/hansmi/prometheus-lvm-exporter ?

insatomcat:
It mostly pays off with thin provisioning (pool data/metadata usage) and snapshots, which I'm not sure is recommended on Seapath. With plain LVs, what's actually missing from node_exporter is VG free space and LV health attributes.
That's a few lines in a textfile collector (`vgs --readonly --units b --nosuffix -o vg_name,vg_free` on a systemd timer). Not fewer moving parts than the exporter, but `vgs` then runs off the scrape path: if a PV stops responding, the timer job hangs instead of the scrape, and we alert on `node_textfile_mtime_seconds`.
Either way, let's not grow this PR: I'd like the review and merge to stay quick. Let's do a follow-up PR for this one.

dupremathieu (approved):

PR #961, opened 2026-06-25
ci: rework how testing is done on Yocto

youennlj-sfl:
- one report PDF per machine (Aaeon, Welotec)
- removed "get_system_info" Ansible role: the CI will use Ansible facts to create the "System Information" report page
- made a reusable `_tests.yml` workflow that can be used for other distributions as well (will be done in future PRs)


youennlj-sfl:
Should be ready to review now.

One question left: the generated PDF are named e.g. `
SEAPATH Yocto Welotec RSAPC Mk2 Tests Report.pdf `. I got a comment that people prefer not to have spaces in file names (mainly because they are less easy to work with in CLIs).
In this case, since it's a very user-facing file that is not made to be manipulated via the CLI, I don't think it's worth introducing code to replace spaces with e.g. underscores. I can be wrong though, please tell me.

.github/test-report/cyclictest.adoc.j2
```diff
@@ -0,0 +1,49 @@
+* Purpose of the test
+  ** Employ *cyclictest* to measure real-time system latency
+    *** Run on isolated CPUs
+    *** Determine worst-case latencies
```
eroussy: Remove this description paragraph. It was not a so good idea after all

.github/test-report/latency_tests.adoc.j2
```diff
@@ -0,0 +1,10 @@
+* Purpose of the test
```
eroussy: Remove this "purpose of the test" line, and just align the two next lines with the rest ( only one `*`)

.github/workflows/_tests.yml
```diff
@@ -0,0 +1,340 @@
...
+            --extra-vars "cukinia_test_prefix=/results" \
+            --extra-vars "cyclictest_result_folder=/results" \
+            --extra-vars "cyclictest_duration=${{ env.CYCLICTEST_DURATION }}" \
+
+          # Check for kernel backtrace error. This is a random error so it must not
+          # stop the CI but just display a warning
+          # See https://github.com/seapath/ansible/issues/164
+          if grep '<failure' ${{ env.RESULTS_FOLDER }}/*.xml | grep -q '00080'; then
+            echo "::warning title=Hypervisor::kernel back trace detected, see https://github.com/seapath/ansible/issues/164"
+          fi
```
eroussy, line 142: Note to self: I think this was an error on the old debian 11 servers and don't impact the Welotec or AAEON machines. We should disable it in the future

.github/workflows/ci-yocto.yml
```diff
@@ -35,8 +35,12 @@ jobs:
           - name: Aaeon ARM I.MX8P-2G
             hyp-inventories: >-
               --inventory inventories_private/ci_yocto_standalone_aaeon.yaml
-    uses: ./.github/workflows/_tests.yml
     name: Run tests on ${{ matrix.name }}
+    concurrency:
+      group: target-${{ matrix.name }}
+      queue: max
+      cancel-in-progress: false
+    uses: ./.github/workflows/_tests.yml
```
eroussy, line 38: This commit is probably not useful, because we will not use mutliple runners for now and we only have one publisher, so we cannot have concurrent access to it

eroussy (changes requested):
Really cool. Small changes and we are good to go

eroussy (approved):

PR #964, opened 2026-06-29
ci: reworked tests on Debian

youennlj-sfl:
Uses the foundations laid out in #961.

Key points:
- one report PDF per setup (cluster vs standalone)


youennlj-sfl:
Branch is on the wrong repository - closing this to reopen later

PR #965, opened 2026-06-29
ci: reworked tests on Debian

youennlj-sfl:
Uses the foundations laid out in https://github.com/seapath/ansible/pull/961.

Key points:
- one report PDF per setup (cluster vs standalone)


.github/workflows/_tests.yml
```diff
@@ -82,7 +89,14 @@ jobs:
           cqfd -b prepare
 
           mkdir ${{ env.RESULTS_FOLDER }}
-          echo "CQFD_EXTRA_RUN_ARGS=-v ${{ env.RESULTS_FOLDER }}:/results" >> "$GITHUB_ENV"
+          CQFD_EXTRA_RUN_ARGS="-v ${{ env.RESULTS_FOLDER }}:/results"
+
+          if [[ "${{ inputs.ansible_ssh_keyfile }}" != '' ]]; then
```
emontmas: `if [[ -n "${{ inputs.ansible_ssh_keyfile }}" ]]` is more concise

Also as a general remark, for coherency, try to keep using either `[` or `[[` in your file, but not both.
sh's `[` and bash's `[[` don't exactly have the same syntax rules, and mixing them is taking a risk to miss quoting a variable or mixing syntaxes

.github/workflows/_tests.yml
```diff
@@ -82,7 +89,14 @@ jobs:
           cqfd -b prepare
 
           mkdir ${{ env.RESULTS_FOLDER }}
-          echo "CQFD_EXTRA_RUN_ARGS=-v ${{ env.RESULTS_FOLDER }}:/results" >> "$GITHUB_ENV"
+          CQFD_EXTRA_RUN_ARGS="-v ${{ env.RESULTS_FOLDER }}:/results"
```
emontmas: Maybe also append the variable here instead of replacing its value? `CQFD_EXTRA_RUN_ARGS="$CQFD_EXTRA_RUN_ARGS -v ${{ env.RESULTS_FOLDER }}:/results"`

So that we don't need to change again this line if in the future:
* We add a task before that again append something to `CQFD_EXTRA_RUN_ARGS`
* We need to inject some extra vars into `CQFD_EXTRA_RUN_ARGS` from the workflow global environment

.github/workflows/_tests.yml
```diff
@@ -388,6 +395,15 @@ jobs:
       - name: Generate test report
         working-directory: test-report-pdf
         run: |
+          compliance_matrices=()
+          if [[ ${{ inputs.compliance_matrices_folder }} != '' ]]; then
```
emontmas: `if [[ -n "${{ inputs.compliance_matrices_folder }}" ]]` is more concise
Also as mentioned be wary of inconsistencies. Here you didn't quoted your path. If the path has a whitespace it will be OK with bash's `[[` but may lead to an error if someone one day decide to change it to sh's `[`

.github/workflows/_tests.yml
```diff
@@ -195,6 +203,12 @@ jobs:
             ${{ inputs.hyp-inventories }} \
             ${{ inputs.vms-inventories }}
 
+          if [[ "${{ inputs.vms-configure-playbooks }}" != '' ]]; then
```
emontmas: if [[ -n "${{ inputs.vms-configure-playbooks }}" ]] is more concise

eroussy (dismissed):

.github/workflows/_tests.yml
emontmas, line 485: Improvement propositions for Github action logs: make the failings steps that triggered this "fail" step to output an error message that is then printed by this "fail" step.
That way, it is easier when debugging a failed workflow to now which step caused the job failure.
Otherwise, the only thing that we see is the "Fail if needed" step in red with `exit 1` as reason

emontmas (changes requested):

.github/workflows/_tests.yml
youennlj-sfl: Whenever the FAILED env variable is set to true by a previous step, an error message is already printed by the same step. There is no need to put other error messages in this end step, whose only task is to set the "failing" status of the whole workflow.

emontmas: Agree, but this means you have to open the previous tasks, and search for the specific error message, which you could miss in the logs
An error message indicating put on the failed step from the POV of the workflow would be clearer in my opinion
But anyway this can be done later

emontmas (approved):

.github/workflows/_tests.yml
youennlj-sfl: No, they are shown as annotations so you can directly see why it fails and click on it to jump to the appropriate step and logs. See https://github.com/seapath/ansible/actions/runs/28585833652/job/84758322730?pr=965

eroussy (approved):

.github/workflows/_tests.yml
emontmas: Oh ok I see, you kind of have to know its there (the annotation panel is fold by default) but ok

youennlj-sfl: It's shown on the frontpage of the workflow when you arrive from the Actions tab.
To me it's a good compromise since it does not require to duplicate the information about which test failed, which would introduce more complexity in the workflow (which is already quite complex).

PR #966, opened 2026-06-30
cyclictest: add variables to set CPU affinity and priority

Obliflo:
Add variables to set the CPU affinity and priority of cyclictest's threads. This replaces `-S` by `-a <cpu-set> -t` to select which core is used.

They are controlled by the cyclictest_affinity and cyclictest_priority variables.
cyclictest_affinity can take two kind of values:
- "smp" : use `-S`, the default
- "cpu-set" : use `-a <cpu-set> -t`


emontmas:
PR is reopened here: https://github.com/seapath/ansible/pull/989

PR #968, opened 2026-07-06
Add SEAPATH SLES Ansible CI

emontmas:
Adds a CI SEAPATH SLES workflow to test Ansible configuration and performances with SEAPATH SLES.
For now, latency tests are not set.

Also add a workflow parameter to specify a branch to checkout for the inventory repository. This can be useful for debugging.

Also adapt some SLES cukinia tests following recent changes in image generation.
Note that the retbleed CPU vulnerability is explicitly skipped to prevent failing the CI as it needs to be discussed how to handle it in SEAPATH SLES.


.github/workflows/ci-yocto.yml
```diff
@@ -40,6 +40,7 @@ jobs:
       compliance_matrices_folder: yocto
       runner-label: "runner-SFL"
       inventories-repository: "git@github.com:seapath-ci/ci-seapath-yocto-inventories.git"
+      inventories-repository-branch: "add-inventories"
```
eroussy: Remove this line

eroussy (changes requested):
Can you remove the ci_yocto reboot on usb role ?
As it is now not only used for yocto

.github/workflows/ci-yocto.yml
emontmas: Done

emontmas:
> Can you remove the ci_yocto reboot on usb role ? As it is now not only used for yocto

@eroussy Done.
Note that I would prefer not doing any additional change in CI roles/playbooks here as this PR is already big enough

eroussy (dismissed):

emontmas:
@eroussy Some linting errors were catch by the CI, I fixed them

eroussy (approved):

PR #969, opened 2026-07-06
workflows/ci-sles.yml: add file

emontmas:
Add a dummy SEAPATH SLES CI workflow file so that it is recognized and added to the list of actions in Github. This is necessary to then run the workflow for test during development.


eroussy (approved):

eroussy:
Merging without CI as this does not modify any code

PR #970, opened 2026-07-06
ceph: explicit per-node OSD disk list, cephadm-only deployment

insatomcat:
## Motivation

The Ceph roles accumulated several layers of history: the original
"dedicated disk" approach, then the "reuse lv_ceph if it exists" logic with
configurable names through `lvm_volumes`, then automatic LV creation and
resizing. An audit of the roles showed that most of these combinations no
longer worked: custom LV names were broken by a hardcoded `vg_ceph/lv_ceph`
zap and spec, the whole ceph-ansible path was dead code, and adding a disk
to a node that already carries OSDs was impossible by construction (OSD
enrollment was decided per host, not per device).

This PR reduces the model to one explicit contract, aligned with what Ceph
recommends: **the inventory declares, per node, the list of block devices
that belong to Ceph (`ceph_osd_disks`), one OSD per device**. Everything
else follows from it.

## What is supported after this PR

- `ceph_osd_disks` host variable: list of paths, whole disks recommended
  (`/dev/disk/by-path/...`, cephadm creates its own LVM layer), pre-existing
  logical volumes accepted for shared-disk setups. `ceph_osd_disk`
  (singular) is kept as a shorthand for a one-element list.
- Growing a cluster = adding a path to `ceph_osd_disks` and re-running the
  playbook, including on nodes that already carry OSDs.
- Custom OSD layouts (separate WAL/DB, encryption, filters) via
  `cephadm_spec_path`: the playbook only ever wipes devices declared in
  `ceph_osd_disks`, the custom spec describes the rest.
- `lvm_volumes` is deprecated but still functional as a frozen compatibility
  shim: it creates the described partition/VG/LV as before, then feeds the
  resulting LV path into `ceph_osd_disks`. Existing CI inventories run
  unchanged (with a deprecation warning).

Dropped: OSD resize (`ceph_expansion_vg`/`ceph_expansion_lv`), automatic
`vg_ceph/lv_ceph` creation on whole disks, the ceph-ansible code paths and
the `is_using_cephadm`/`force_cephadm` switch, and the `mons`/`osds`/
`clients` inventory groups inherited from ceph-ansible.

## The safety rule for wiping devices

Whether a device must be wiped is no longer tied to host state or to user
flags. A single rule applies: **a device is kept untouched if and only if it
hosts an OSD claimed by the running cluster** — the OSD uuid stored in the
LVM tags written by ceph-volume is present in the cluster OSD map
(`ceph osd dump`). Anything else declared in `ceph_osd_disks` belongs to
Ceph and is zapped (logical volumes in place, whole disks with `--destroy`),
after being listed in the playbook output.

This one rule covers the whole lifecycle without user input:

- re-running the playbook on a live cluster never touches existing OSDs;
- reinstalling from scratch (fresh fsid, empty OSD map) automatically wipes
  and re-enrolls the volumes of the previous installation — the standard CI
  scenario;
- a replaced machine, whose OSD was removed from the map, has its volume
  automatically re-enrolled.

In addition, `ceph_prepare_installation` validates the declared devices
before anything destructive runs: a path that does not exist, or a device
holding mounted filesystems or swap (e.g. a mistyped system disk), fails the
play instead of being wiped.

## Commits

1. **ceph: remove ceph-ansible leftovers, cephadm is the only deployment
   path** — removes the dead ceph-ansible branches, the
   `is_using_cephadm`/`force_cephadm` switch and the vendored-directory
   references. No behavior change for cephadm deployments.
2. **ceph: remove OSD resize support** — deletes `ceph_expansion_vg`,
   `ceph_expansion_lv` and their plays.
3. **cephadm: decide OSD zapping per volume instead of per host** —
   replaces the per-host "needs OSDs" detection and the hardcoded
   `vg_ceph/lv_ceph` zap with the claim check described above.
4. **ceph: give Ceph an explicit list of disks per node** — introduces
   `ceph_osd_disks`, the validation step, the per-node generated spec
   (service id = hostname), the type-aware zap, the `lvm_volumes` shim, and
   extends the smartctl sudoers rule to every declared disk.
5. **inventory: retire the ceph-ansible mons/osds/clients groups** — plays
   target `cluster_machines`; OSD nodes are the ones declaring
   `ceph_osd_disks`. The example inventory keeps only the live Ceph
   variables (`cephadm_network`, `ceph_conf_overrides`).
6. **playbooks: clean up the machine-replacement playbooks** — deletes
   `replace_machine_find_online_ceph_machine.yaml` and
   `replace_machine_shrink_pacemaker.yaml`, orphaned since the
   ceph-ansible shrink playbooks were removed, and renames
   `replace_machine_remove_machine_cephadm.yaml` to
   `cluster_remove_machine.yaml` (cephadm is the only path, no need to
   qualify it).

Each commit leaves the repository consistent and can be reviewed (or
reverted) independently.

## Backward compatibility

- Single-disk inventories using `ceph_osd_disk`: unchanged, no edit needed.
- Shared-disk inventories using `lvm_volumes`: unchanged, deprecation
  warning; migration path is to create the LV at provisioning time and list
  it in `ceph_osd_disks`.
- Inventories still defining the `mons`/`osds`/`clients` groups: harmless,
  the groups are simply no longer consumed (variables attached to them still
  propagate to their member hosts).
- Existing clusters deployed on `vg_ceph/lv_ceph`: safe to re-run, the
  volumes are claimed and left untouched. The newly applied spec declares
  per-hostname OSD services next to the legacy `osd.osdN` ones in
  `ceph orch ls`; full migration to the whole-disk layout is a documented
  manual procedure (see the cephadm role README).
- `force_cephadm: false` no longer exists; not deploying Ceph on a node is
  expressed by not declaring `ceph_osd_disks`.

## Validation

- `ansible-lint` (production profile) passes on all modified roles and
  playbooks.
- The claim-check logic was validated against realistic
  `ceph-volume lvm list` data for the four lifecycle scenarios (live
  re-run, scratch reinstall, machine replacement, fresh volume).
- The generated OSD spec was rendered against multi-disk, LV-based and
  disk-less nodes and checked to be valid YAML with the expected services.
- The example inventory loads with `ansible-inventory` and resolves
  `ceph_osd_disks`/`cephadm_network`/`ceph_conf_overrides` on all nodes.
- Not yet exercised on a real cluster — the integration CI should be the
  gate for this PR, in particular the `lvm_volumes` shim on the shared-disk
  runners.

----

also, some fixes:

bump cukinia

network_buildhosts: match /etc/hosts entries by inventory hostname
The previous regexp matched on the hostname column, which could miss
existing lines and create duplicate entries when hostname differs from
the inventory name.

cephadm: detect cluster membership with cephadm ls
Use local daemon listing instead of ceph -s so a host is only considered
part of the cluster when it actually runs Ceph services.

cephadm: remove ceph-common logrotate config to avoid duplicate entry
ceph-common installs a generic /var/log/ceph/*.log logrotate config
that matches cephadm.log, which cephadm already manages via its own
/etc/logrotate.d/cephadm. logrotate then fails with "duplicate log
entry for /var/log/ceph/cephadm.log".
The distro-specific path for that conflicting file was previously
tracked as dead code (cephadm_logrorateceph_path, only used by a
replace task removed in https://github.com/seapath/ansible/commit/22e302c3887975537a95145b1289f7dc46716e6a). Move and rename it to
cephadm_install_logrotateceph_path, following the existing
{Distro}_cephadm_install.yml convention, and remove it right after
ceph-common is installed.

ci_restoredd: sync before forced reboot so RESTORE_PLZ marker hits the ESP
systemctl reboot --force skips filesystem unmount, so the freshly touched
/boot/efi/RESTORE_PLZ could still sit in the page cache and be lost, the
ESP being FAT32 and not mounted sync. Run sync in the reboot command so
the marker is on disk before the reset.


dupremathieu (approved):
Great!!!

PR #971, opened 2026-07-07
Make RX queue size configurable for bridge interfaces

viccie30:
When using a bridge to receive sampled values, the default queue size is too small, leading to bursty behavior.

### DIFF
--- a/templates/vm/guest.xml.j2
+++ b/templates/vm/guest.xml.j2
@@ -170,6 +170,9 @@
             <source bridge="{{ bridge.name }}"/>
             <mac address="{{ bridge.mac_address }}"/>
             <model type="virtio"/>
+{%         if bridge.rx_queue_size is defined %}
+            <driver rx_queue_size="{{ bridge.rx_queue_size }}"/>
+{%         endif %}
 {%         if bridge.type is defined %}
                 <virtualport type='{{ bridge.type }}'/>
 {%         endif %}
@@ -187,6 +190,9 @@
             <mac address="{{ bridge.mac_address }}"/>
             <target dev="{{ bridge.ovs_port }}" managed="no"/>
             <model type="virtio"/>
+{%         if bridge.rx_queue_size is defined %}
+            <driver rx_queue_size="{{ bridge.rx_queue_size }}"/>
+{%         endif %}
         </interface>
 {%     endfor %}
 {% endif %}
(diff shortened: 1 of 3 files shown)


templates/vm/guest.xml.j2
```diff
@@ -170,6 +170,9 @@
             <source bridge="{{ bridge.name }}"/>
             <mac address="{{ bridge.mac_address }}"/>
             <model type="virtio"/>
+{%         if vm.rx_queue_size is defined %}
+            <driver rx_queue_size='{{ vm.rx_queue_size }}'/>
```
insatomcat: This reads `vm.rx_queue_size` (VM-level), but both READMEs document `rx_queue_size` as a per-bridge field inside the `bridges` list, next to `name` and `mac_address`. A user following the documentation will set it on a bridge entry and it will be silently ignored — no error, just the default ring size and the burstiness they were trying to fix.

Since the docs already describe the per-interface form (and it is the more flexible one), I would make the template match them:

```jinja
{%         if bridge.rx_queue_size is defined %}
            <driver rx_queue_size='{{ bridge.rx_queue_size }}'/>
{%         endif %}
```

insatomcat (changes requested):
Thanks, making the virtio RX ring size tunable definitely makes sense for sampled values reception. One inconsistency to fix before merge, see inline comment.

Side note, not blocking: the same need exists for OVS ports (`type="ethernet"` interfaces just below), which is the other common path for receiving SV. Fine to keep that out of scope here, but worth a follow-up.

viccie30:
I've made both changes you've requested, they make sense! Thanks for catching my documentation error.

insatomcat:
Thanks for the quick update! Both fixes look good: the template now matches the documented per-bridge form, and extending it to OVS ports covers the other common path for SV reception.

Two small leftovers:
- The READMEs only document `rx_queue_size` under the `bridges` section, but the template now also supports it on `ovs` entries. Could you add the same line to the `ovs` table in both `roles/deploy_vms_cluster/README.md` and `roles/deploy_vms_standalone/README.md`? Otherwise the OVS variant works but is undiscoverable.
- While you're in that table: `mad_address` in the `ovs` section is a pre-existing typo for `mac_address` — feel free to fix it in passing :)

Also, take it or leave it: the new `<driver rx_queue_size='...'/>` uses single quotes where the surrounding lines use double quotes. The template already mixes both styles, so this is purely cosmetic.

viccie30:
> Thanks for the quick update! Both fixes look good: the template now matches the documented per-bridge form, and extending it to OVS ports covers the other common path for SV reception.
> 
> Two small leftovers:
> 
> * The READMEs only document `rx_queue_size` under the `bridges` section, but the template now also supports it on `ovs` entries. Could you add the same line to the `ovs` table in both `roles/deploy_vms_cluster/README.md` and `roles/deploy_vms_standalone/README.md`? Otherwise the OVS variant works but is undiscoverable.

Done
> * While you're in that table: `mad_address` in the `ovs` section is a pre-existing typo for `mac_address` — feel free to fix it in passing :)

Done
> 
> Also, take it or leave it: the new `<driver rx_queue_size='...'/>` uses single quotes where the surrounding lines use double quotes. The template already mixes both styles, so this is purely cosmetic.

Done. I agree styles should match.

I also fixed a wonky pipe at the end of an earlier row in the table.

insatomcat (approved):

PR #977, opened 2026-07-17
cephadm_install: move distro vars into the role

insatomcat:
The role's per-distro variables (uid/gid, logrotate path) were only ever consumed by cephadm_install itself, so loading them from /vars/{Distro}_cephadm_install.yml at playbook level added no value over the standard in-role vars/ convention already used by roles like snmp and timemaster.
Move them to roles/cephadm_install/vars/{Distro}.yml and include_vars from within the role's own tasks/main.yml, dropping the now-unneeded include_vars step from cluster_setup_cephadm.yaml.


insatomcat:
PR merged into #983

PR #979, opened 2026-07-18
cephadm_install: detect the ceph image version from local images

insatomcat:
build_debian_iso now embeds the latest cephadm container image, so the hardcoded cephadm_release default can diverge from the image actually present on the nodes and break the deployment.

When cephadm_release is not set in the inventory, detect the version of the quay.io/ceph/ceph image already present on the nodes and use it for the whole deployment. The detection fails if a node holds several ceph image versions or if two nodes hold different versions. When no image is found anywhere, fall back to the default version. Setting cephadm_release in the inventory disables the detection and pins the version.


insatomcat:
PR merged into #983

PR #981, opened 2026-07-21
cephadm_install/README.md: fix copy-paste in registryurl description

insatomcat:
cephadm_install_registryurl was documented as "Name of the ceph release for repo install (fallback: cephadm_release_name)", copied from the row above. It is the image used to start the local Podman registry and has no fallback variable.


insatomcat:
PR merged into #983

PR #982, opened 2026-07-21
cephadm_install: derive the Ceph release name from the release version

insatomcat:
"cephadm add-repo" takes a release name while the cephadm download URL takes a version number. Both were pinned separately, so bumping cephadm_install_release alone silently left the repo on the previous release, mixing a binary from one Ceph version with packages from another.

Derive the name from the major version through a lookup table, and assert before add-repo that a name is known, so an unlisted version fails with an explicit message instead of installing the wrong repo. Setting cephadm_install_release_name explicitly still overrides the derivation.


insatomcat:
PR merged into #983

PR #983, opened 2026-07-22
roles/cephadm_install: fixes

insatomcat:
**cephadm_install: move distro vars into the role**

The role's per-distro variables (uid/gid, logrotate path) were only
ever consumed by cephadm_install itself, so loading them from
/vars/{Distro}_cephadm_install.yml at playbook level added no value
over the standard in-role vars/ convention already used by roles like
snmp and timemaster. Move them to roles/cephadm_install/vars/{Distro}.yml
and include_vars from within the role's own tasks/main.yml, dropping
the now-unneeded include_vars step from cluster_setup_cephadm.yaml.

----
**cephadm_install: detect the ceph image version from local images**

build_debian_iso now embeds the latest cephadm container image, so the
hardcoded cephadm_release default can diverge from the image actually
present on the nodes and break the deployment.
When cephadm_release is not set in the inventory, detect the version of
the quay.io/ceph/ceph image already present on the nodes and use it for
the whole deployment. The detection fails if a node holds several ceph
image versions or if two nodes hold different versions. When no image is
found anywhere, fall back to the default version. Setting
cephadm_release in the inventory disables the detection and pins the
version.

----
**cephadm_install/README.md: fix copy-paste in registryurl description**

cephadm_install_registryurl was documented as "Name of the ceph release
for repo install (fallback: cephadm_release_name)", copied from the row
above. It is the image used to start the local Podman registry and has
no fallback variable.

----
**cephadm_install: derive the Ceph release name from the release version**

"cephadm add-repo" takes a release name while the cephadm download URL
takes a version number. Both were pinned separately, so bumping
cephadm_install_release alone silently left the repo on the previous
release, mixing a binary from one Ceph version with packages from
another.
Derive the name from the major version through a lookup table, and
assert before add-repo that a name is known, so an unlisted version
fails with an explicit message instead of installing the wrong repo.
Setting cephadm_install_release_name explicitly still overrides the
derivation.


roles/cephadm_install/README.md
eroussy: Remove default values

roles/cephadm_install/defaults/main.yml
```diff
@@ -4,7 +4,19 @@
...
-cephadm_install_release_name: "{{ cephadm_release_name | default('tentacle') }}"
+# "cephadm add-repo" takes a release name while the binary download URL takes a
+# version number, so both have to designate the same Ceph release. Derive the
+# name from the major version instead of pinning it separately.
+cephadm_install_release_names:
+  "17": "quincy"
+  "18": "reef"
+  "19": "squid"
+  "20": "tentacle"
+  "21": "umbrella"
```
eroussy: I don't like the idea to put that as default of a variable. The user will never need to overwrite that IMO.
I would rather put that as a set_fact and remove it from here and from the README.

eroussy (changes requested):

roles/cephadm_install/README.md
insatomcat: Done, the row is removed from the README along with the variable itself (see the other comment).

roles/cephadm_install/defaults/main.yml
insatomcat: Agreed, this is internal data of the role, not something a user should configure. I moved the table to the role's vars/main.yml rather than a set_fact though: it ends up just as non-overridable (role vars have high precedence), but it stays declarative and doesn't add a task run on every node. It is gone from defaults/ and from the README; cephadm_install_release_name remains the user-facing override, as the assert message suggests.

roles/cephadm_install/README.md
```diff
@@ -2,20 +2,33 @@
...
+- If no image is found on any node (image pulled at deploy time), the default version below is used and a warning is printed.
+
+The detected version is set as a host fact, so it also applies to the `cephadm` role (bootstrap, `--image` flags) in later plays of the same run.
+
 ## Role Variables
 
 | Variable                          | Required | Type    | Default                        | Comments                                                                                   |
 |-----------------------------------|----------|---------|--------------------------------|--------------------------------------------------------------------------------------------|
 | seapath_distro                    | No       | String  | Not set                        | SEAPATH distribution                                                                       |
 | cephadm_install_release           | No       | String  | "20.2.0"                       | Version of the cephadm binary to download (fallback: `cephadm_release`)                    |
```
eroussy: Still need to remove the defaults value of this

eroussy (changes requested):

roles/cephadm_install/README.md
insatomcat: Right, since the detection commit, "20.2.0" is not really the default anymore: the default behavior is to detect the version from the local ceph image, and the hardcoded version is only a last-resort fallback when no image is found anywhere. The Default cell now says "Detected from the local ceph image" and the hardcoded number is gone from the README entirely; the fallback itself stays in defaults/main.yml for the no-image case, as described in the detection section.

eroussy (approved):

PR #985, opened 2026-07-23
Fix cluster disruptions caused by tuned profile application on hypervisors

insatomcat:
Applying (or re-applying) the seapath-rt-host tuned profile on a live
  hypervisor was disruptive to the cluster: several plugins inherited from
  the realtime-virtual-host base profile perform actions that stall CPUs,
  reset NICs or migrate threads at apply time. On RT kernels these stalls
  are long enough to break Corosync heartbeat forwarding, causing token
  loss, quorum loss, and Pacemaker stopping cluster resources (VMs,
  quadlets). Re-applying the profile could also crash running guests by
  moving pinned QEMU vCPU threads off their isolated cores.

  This PR makes tuned profile application safe on a running cluster node.

  ### Changes

  - **Disable irqbalance system-wide** (`387b8b8`): irqbalance fights
    nic-irq-monitor.sh, and its restart during profile application causes
    ~1s of IRQ redistribution that breaks Corosync heartbeats. Masked on
    all distros; tuned's [irqbalance] plugin suppressed.

  - **Make tuned restart/profile application conditional** (`273836e`):
    tuned.service was always restarted and the profile always re-applied,
    even when nothing changed. Both are now conditional on actual
    configuration changes or the profile not being active.

  - **Disable the scheduler plugin's process-affinity management**
    (`eb055cd`): the base profile moved processes off isolated cores,
    un-pinning QEMU vCPU threads and killing containers pinned via a
    cgroup-level CPUAffinity (empty CPU intersection). Redundant with
    isolcpus= anyway.

  - **Prevent workqueue migration during profile reload** (`91a2702`):
    writing workqueue cpumasks migrated all unbound workqueue threads
    (~3s), starving Corosync. Overridden to a no-op mask; isolcpus=
    already keeps new tasks off isolated cores.

  - **Fix profile reload rollback disrupting Corosync** (`68c0fe7`):
    re-applying an active profile runs a rollback phase whose [net]
    plugin resets NIC channels (~1s NIC reset on the Corosync ring).
    [net] is now disabled in the profile, and the Ansible task restarts
    the tuned service instead of running tuned-adm, flushing rollback
    state.

  - **Disable the scheduler plugin entirely, restore rcuc priority**
    (`f3be359`): the plugin's perf_event mmap() calls across all CPUs
    cause scheduling stalls on RT kernels; simultaneous application on
    all nodes cascades into quorum loss. The rcuc FIFO:10 priority tuning
    it provided is preserved via a custom script.sh in the profile.

  ### Result

  The seapath-rt-host profile can now be applied or re-applied on a node
  with running VMs and cluster resources without losing quorum, resetting
  NICs, or disturbing pinned RT threads. Process isolation remains fully
  enforced by the kernel cmdline (isolcpus=managed_irq, nohz_full,
  rcu_nocbs), which tuned's bootloader plugin still configures.


eroussy:
I will not take the time to review the entire PR now, but just one short question : 
Why not remove irqbalance package from SEAPATH entirely ?

insatomcat:
Good question, but I think it mixes two concerns.

This PR isn't doing package management: it masks `irqbalance.service` (runtime state) and drops tuned's `[irqbalance]` plugin (config), both squarely Ansible's job.

Dropping the package itself is a fine complementary idea, but the right place is the source (`build_debian_iso` / `meta-seapath`), not here. Having Ansible uninstall packages would be a new pattern for this repo and hurts idempotency (revert, CentOS/Debian differences). The convention is that Ansible configures systems that already ship the right packages, it doesn't add or remove them.

And even with the package gone, I'd keep the mask + plugin change: it's correct whether or not irqbalance is installed, and the actual root cause is the base realtime profile's `[irqbalance]` plugin interaction with tuned, not the package.

So: happy to see it dropped in meta-seapath/build_debian_iso, but I'd keep the mask here.

roles/configure_hypervisor/tasks/tuned.yml
```diff
@@ -20,23 +20,40 @@
...
   ansible.builtin.systemd:
     name: tuned.service
     enabled: true
+    state: started
+
+- name: Restart tuned after config change
+  when: configure_hypervisor_tuned_main_conf.changed
+  ansible.builtin.systemd:
+    name: tuned.service
     state: restarted
```
eroussy, line 61: This should be a handler

roles/configure_hypervisor/tasks/tuned.yml
```diff
@@ -30,31 +30,31 @@
...
-  ansible.builtin.command: tuned-adm profile seapath-rt-host
   when: >
-    isolcpus is defined and
-    (configure_hypervisor_tuned_conf.changed or
-     'seapath-rt-host' not in configure_hypervisor_tuned_active.stdout)
-  changed_when: true
+    configure_hypervisor_tuned_main_conf.changed or
+    (isolcpus is defined and
+     (configure_hypervisor_tuned_conf.changed or
+      configure_hypervisor_tuned_profile_set.changed))
```
eroussy, line 67: This should call a handler

roles/configure_hypervisor/tasks/tuned.yml
```diff
@@ -22,6 +22,15 @@
     mode: "0644"
   register: configure_hypervisor_tuned_conf
 
+- name: Copy seapath-rt-host tuned profile script
+  ansible.builtin.copy:
+    src: tuned-script.sh
+    dest: "{{ configure_hypervisor_tuned_profiles_dir }}/seapath-rt-host/script.sh"
```
eroussy, line 28: give a meaningfull name to this script

roles/configure_hypervisor/files/tuned-script.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
eroussy: Give a meaningfull name to this script

roles/configure_hypervisor/files/tuned-script.sh
```diff
@@ -0,0 +1,26 @@
...
+# Copyright (C) 2026 RTE
+# SPDX-License-Identifier: Apache-2.0
```
eroussy: Add comment here to explain the relation to realtime-virtual-host script

eroussy (changes requested):
I don't see a point to this first commit
IMO, you should directly remove irqbalance in  build_debian_iso and remove all irqbalance reference of ansible in this PR

insatomcat:
> I don't see a point to this first commit IMO, you should directly remove irqbalance in build_debian_iso and remove all irqbalance reference of ansible in this PR

I agree the package should go away in build_debian_iso, and I am happy to open that separately. But I do not think it lets us drop the handling here, for a reason that is independent of what the image ships.

Ansible is decoupled from build_debian_iso. It has to be: a playbook is declarative, so the state it converges to must not depend on an external element it does not control. The image is one such element, and an admin installing irqbalance by hand after deployment on a running node is another. In both cases the correct outcome is the same: after configure_hypervisor has run, irqbalance is not fighting nic-irq-monitor.sh for NIC IRQ affinity. If Ansible only works because the package happens to be absent, we have not configured anything, we have just been lucky.

So Ansible has to know how to handle irqbalance if it is there. That is exactly what Stop and mask irqbalance.service does, and masking is the right primitive: it is idempotent, it is a no-op if the unit does not exist, and it survives a later apt install irqbalance.

Two more practical points:
- The commit already does most of what you are asking. It deletes tasks/centos/irqbalance.yml (the IRQBALANCE_BANNED_CPULIST tuning) and the Restart irqbalance handler. What is left is 10 lines of masking, so the discussion is just on those 10 lines rather than on the commit.
- [irqbalance] enabled=0 in tuned.conf.j2 is needed regardless. It stops tuned from stopping and restarting the service at profile apply time, which is one of the ~1s stalls that break Corosync, and it applies to CentOS and SLES too, where removing the package from build_debian_iso changes nothing.

Also worth noting for sequencing: irqbalance is currently listed as an expected package in packages_tests.d/debian.conf.j2 and sles.conf.j2, and as an expected unit in systemd.conf.j2. Removing it from the images means updating those cukinia tests in the same window, which is another argument for doing it as its own change rather than folding it into this PR.

roles/configure_hypervisor/tasks/tuned.yml
insatomcat: You are right that this is the idiomatic pattern, and it is what the rest of this repo does (Restart sysfsutils, Start new slices). I went the explicit way here on purpose, because restarting tuned is not an ordinary service reload in this role.

1. Ordering. tuned.yml ends with tuned-adm off when isolcpus is not defined. Handlers are flushed at the end of the play, so the handler would fire after that and bring the profile back up. The restart has to happen at a defined point inside the role, not at an unrelated point later.
2. Silent divergence on failure. There is no force_handlers anywhere in this repo. If any task later in the play fails, the handler never runs, and on the next run the files are already up to date, so nothing is changed and the restart never happens again. The node would then have the new profile on disk and the old one live, indefinitely, with nothing reporting it. Given that this profile drives the RT isolation of a cluster node, that silent divergence is precisely the class of problem this PR exists to fix.
3. meta: flush_handlers is not a clean fix. It flushes all pending handlers of the play, so Restart sysfsutils (from sriov.yml) and Start new slices (from slices.yml, immediately before) would move from end of play into the middle of the role. Trading a local when: for a global side effect on unrelated tasks is a net loss.
4. The condition is not "a file changed". tuned.conf and script.sh are written unconditionally so the on-disk state is complete, but they must only trigger a restart when isolcpus is defined. notify has no conditional form, so expressing this would mean putting when: isolcpus is defined on tasks that must run in both cases, which changes the resulting state on disk.
5. Last, the restart is the disruptive action of this whole PR. I would rather have it appear explicitly in the task list, at a point I control, than deferred to the end of the play.

So I really thought that in this particular usecase, not using the "normal" way was better.

roles/configure_hypervisor/tasks/tuned.yml
insatomcat: see above

roles/configure_hypervisor/tasks/tuned.yml
insatomcat: Done, the source file is now files/seapath-rt-host-script.sh instead of files/tuned-script.sh.

I kept the destination as <tuned_profiles_dir>/seapath-rt-host/script.sh though. The file is installed inside the profile directory, which already namespaces it, and script.sh is the name every tuned profile uses for its script plugin (realtime-virtual-host, latency-performance, and the rest all ship script.sh, referenced through ${i:PROFILE_DIR}/script.sh). Renaming it on the target would only make our profile the odd one out for anyone inspecting /usr/lib/tuned/ or /etc/tuned/ on a node, without gaining any clarity, since the path already says seapath-rt-host.

So the meaningful name is on the Ansible side, where the file sits next to 30-watchdog.conf in a flat files/ directory and does need to be self-describing, and the conventional name is on the target side.

roles/configure_hypervisor/files/tuned-script.sh
insatomcat: Good catch, this was the least obvious part of the change. Added a header explaining both halves of it:

```
# Script plugin of the seapath-rt-host tuned profile.
#
# seapath-rt-host includes realtime-virtual-host, and declaring [script] here
# overrides the script inherited from that base profile: its own script.sh is
# no longer executed. Its KVM low latency setup is therefore reproduced below,
# and must be kept in sync with the base profile.
#
# The rcuc priority tuning replaces the [scheduler] group.rcuc rule of the base
# profile, which no longer applies since the scheduler plugin is disabled: its
# perf_event mmap() calls stall CPUs on RT kernels when the profile is applied.
```

The point being that setup_kvm_mod_low_latency here is not an addition on top of the base profile, it is a replacement for it: without it we would silently lose the KVM low latency setup the moment we declare our own [script]. Same for the rcuc chrt, which carries over what `[scheduler] group.rcuc=0:f:10:*:^\[rcuc` used to do before the scheduler plugin was disabled in the other commit.

roles/configure_hypervisor/files/tuned-script.sh
insatomcat: Done, the source file is now files/seapath-rt-host-script.sh instead of files/tuned-script.sh.

I kept the destination as <tuned_profiles_dir>/seapath-rt-host/script.sh though. The file is installed inside the profile directory, which already namespaces it, and script.sh is the name every tuned profile uses for its script plugin (realtime-virtual-host, latency-performance, and the rest all ship script.sh, referenced through ${i:PROFILE_DIR}/script.sh). Renaming it on the target would only make our profile the odd one out for anyone inspecting /usr/lib/tuned/ or /etc/tuned/ on a node, without gaining any clarity, since the path already says seapath-rt-host.

So the meaningful name is on the Ansible side, where the file sits next to 30-watchdog.conf in a flat files/ directory and does need to be self-describing, and the conventional name is on the target side.

eroussy (approved):

PR #1004, opened 2026-08-18
Stage VM disks from local or remote sources via a shared stage_vm_disk role

dahoat-sprecher:
Hello to All,

this supersedes #930 which was not merged as I started extending the functionality. I will close #930 in favor of this PR.

Intent
Let operators provision VMs from disk images hosted on a remote URL, fetched directly by the target node via ansible.builtin.get_url, instead of staging every image from the Ansible control machine. Integrity is guaranteed by a mandatory checksum for remote sources.

What changed compared to #930
- No separate vm_remote_disk variable anymore: vm_disk (and additional_disk entries) simply accept a URL (http/https/ftp/ftps) instead of a local path. URL-shaped entries are fetched on the target, everything else keeps the existing copy path.
- The staging logic moved into a new internal role stage_vm_disk, shared by deploy_vms_cluster and deploy_vms_standalone and used for system disks as well as additional disks.

The stage_vm_disk role
- Validates the disk entry, resolves local vs. remote source (with an optional fallback path from the role's disks directory) and stages the image on the target.
- vm_disk_checksum / inline checksum keys: dict {<algo>: <hex>} with sha256/sha384/sha512; the strongest provided algorithm is used. Required for remote sources; optional for local ones, but if given it is verified after the copy and a mismatching file is removed from the target before the play fails.
- stage_vm_disk_validate_certs allows opting out of TLS certificate validation (e.g. internal mirrors); the mandatory checksum still guarantees integrity.
- A normalize entry point converts mixed additional_disk lists (path strings or dicts with path, checksum keys and an optional label) into a normalized list; the label is reflected in the staged file name (<vm>_data_<index>_<label>.qcow2).
- Comes with a Molecule test suite covering the validation and staging paths.

Changes in the deployment roles
- deploy_vms_cluster: system disk and additional disks are staged through stage_vm_disk; behavior for plain local paths is unchanged.
- deploy_vms_standalone: same migration, plus additional-disk normalization now runs before any destructive step, so an invalid disk entry fails the play before an existing VM is destroyed (force) or any disk is staged. New deploy_vms_standalone_vms_disks_directory default provides the fallback directory analogous to the cluster role.
- templates/vm/guest.xml.j2 consumes the normalized additional-disk file names (standalone_additional_disks).
- READMEs of all touched roles updated accordingly.

All the best,
Daniel


roles/stage_vm_disk/tasks/main.yml
```diff
@@ -0,0 +1,132 @@
...
+    - not stage_vm_disk_source_is_remote
+
+- name: "Remove disk with mismatching checksum for {{ stage_vm_disk_label }}"
+  ansible.builtin.file:
+    path: "{{ stage_vm_disk_destination }}"
+    state: absent
+  when:
+    - stage_vm_disk_checksum_present
+    - not stage_vm_disk_source_is_remote
+    - stage_vm_disk_stat.stat.checksum | default('') != stage_vm_disk_checksum_hash
```
insatomcat, line 120: Blocking: `stat` always returns lowercase hex, and this compares it to the inventory value verbatim, so an uppercase digest is a mismatch. `get_url` normalizes with `re.sub(r'\W+', '', checksum).lower()`, so the very same digest passes for a URL source and fails for a local one.

Tested with `sha256: "1A32...E"`: the role copies the file, removes the destination, then fails with `expected sha256:1A32... got sha256:1a32...`. In `deploy_vms_standalone` the destination is `/var/lib/libvirt/images/<vm>.qcow2`, so a correct digest in the wrong case deletes the VM system disk.

`| trim | lower` on both sides here and in the assert below.

roles/deploy_vms_cluster/tasks/main.yml
```diff
@@ -59,23 +59,27 @@
...
+    - name: "Stage system disk on target for {{ item }}"
+      ansible.builtin.include_role:
+        name: stage_vm_disk
       vars:
-        ansible_remote_tmp: "{{ deploy_vms_cluster_qcow2tmpuploadfolder | default(omit) }}"
+        stage_vm_disk_label: "{{ item }}"
+        stage_vm_disk_destination: "{{ vm_file_dest }}"
+        stage_vm_disk_source: "{{ hostvars[item].vm_disk | default('') }}"
+        stage_vm_disk_fallback_source: "{{ deploy_vms_cluster_vms_disks_directory ~ '/' ~ item ~ '.qcow2' if deploy_vms_cluster_vms_disks_directory | length > 0 else '' }}"
+        stage_vm_disk_checksum: "{{ hostvars[item].vm_disk_checksum | default({}) }}"
```
insatomcat, line 72: Blocking: this `include_role` var outranks the `defaults/main.yml` expression that extracts the checksum keys from `stage_vm_disk_source`. With `vm_disk: {path: /local/x.qcow2, sha256: ...}` the inline checksum is silently dropped and no verification runs at all. I reproduced it with a deliberately wrong `sha256`: the play reports success.

The same shape with a URL fails instead with the misleading "A checksum is required when the disk source is a URL". `stage_vm_disk`'s README documents `stage_vm_disk_source` as a raw inventory entry, so the dict form is reachable.

`| default(omit)`, or merge the two sources rather than overriding. Same line at `roles/deploy_vms_standalone/tasks/main.yml:66`.

roles/deploy_vms_standalone/tasks/main.yml
```diff
@@ -7,6 +7,17 @@
     command: list_vms
   register: deploy_vms_standalone_all_vms
 
+# Runs first so invalid disk entries fail before the existing VM is
+# destroyed (force) or any disk is staged.
+- name: Normalize additional disk file names for the VM templates
```
insatomcat, line 12: The comment promises more than the task delivers. Normalization only validates entry shape (string, or dict with `path`/`label`/`sha*`). The checks that actually fire in practice live in `stage_vm_disk` and run at staging time, well after `Stop VM` and `virsh undefine --nvram`.

Concretely, `vm_disk: "https://.../guest.qcow2"` with `force: true` and no `vm_disk_checksum` leaves the hypervisor with the VM destroyed and the play failed. The system disk gets no pre-destroy validation at all, only the additional disks do.

roles/deploy_vms_standalone/tasks/normalize_additional_disk.yml
```diff
@@ -0,0 +1,30 @@
...
+    stage_vm_disk_label: "{{ _vm_name }}"
+    stage_vm_disk_disks: "{{ hostvars[_vm_name].additional_disk }}"
+
+- name: "Reset additional disk file names for {{ _vm_name }}"
+  ansible.builtin.set_fact:
+    deploy_vms_standalone_additional_disk_files: []
+
+- name: "Derive additional disk file names for {{ _vm_name }}"
+  ansible.builtin.set_fact:
+    deploy_vms_standalone_additional_disk_files: "{{ deploy_vms_standalone_additional_disk_files + [_vm_name ~ '_data_' ~ add_disk_idx ~ ('_' ~ add_disk.label if add_disk.label | length > 0 else '') ~ '.qcow2'] }}"
```
insatomcat, line 19: This renames staged additional disks from `<basename>.qcow2` to `<vm>_data_<idx>[_<label>].qcow2`, which is a breaking change that the remote-fetch feature does not require.

Any user `vm_template` looping over `vm.additional_disk` (the pattern the bundled template used until this PR, and that `roles/deploy_vms_standalone/README.md:98` still recommends) will define a VM pointing at files that no longer exist. And on a `force` redeploy of a VM created by an earlier version, the old `<disk_pool>/<basename>.qcow2` files are never removed: the VM is re-pointed at fresh copies, the guest data is orphaned, and the pool grows by a full image each time.

Could this keep the existing names, or land as an opt-in with a migration note?

insatomcat (commented):
Thanks for this. The feature itself is worth having: fetching on the target removes the per-hypervisor push over SSH from the control machine, it makes the roles usable from AWX where the images are not on the controller, and the mandatory checksum closes a real gap, since VM disks are staged today with no integrity check at all. Folding four duplicated copy paths into one role is a good side effect.

Four points inline before this leaves draft, the first two blocking. I ran `stage_vm_disk` locally to confirm both.

Three minor ones not worth an inline comment:

- `stage_vm_disk_remote_tmp` has no default, and `omit` inside `vars:` is not stripped the way it is for module args, so `ansible_remote_tmp` resolves to an empty string for any caller that does not pass it. Your molecule scenario does not.
- the destination derived in `roles/stage_vm_disk/defaults/main.yml` keeps the URL query string, so `https://ex.com/a.qcow2?token=xyz` stages a file literally named `a.qcow2?token=xyz`. Signed-URL mirrors are the use case this PR targets.
- `roles/deploy_vms_standalone/README.md:98` still says custom templates may keep consuming the raw `additional_disk` entries. That is no longer true, see the last inline comment.

PR #1010, opened 2026-08-25
inventories/examples: add v1.2.0 to v2.x migration guide

dupremathieu:
## Description

Add `inventories/examples/MIGRATION-v1.2.0-to-v2.x.md`, a guide explaining how to update existing inventories when migrating from SEAPATH ansible v1.2.0 to v2.x (main).

It covers, for each example inventory:
- groups to remove (ceph-ansible `mons`/`osds`/`clients` and the empty placeholder groups)
- variables to change (`livemigration_user` -> `libvirtadmin`)
- new mandatory variables (per-host `ceph_osd_disks`, `cephadm_network` + `ceph_conf_overrides` in `cluster_machines:vars`, `live_migration` for VMs)
- new optional variables (`ptp_domain_number`, `grub_password`, `deploy_cephfs`, `ansible_ssh_private_key_file`)

### DIFF
--- a/inventories/examples/MIGRATION-v1.2.0-to-v2.x.md
+++ b/inventories/examples/MIGRATION-v1.2.0-to-v2.x.md
@@ -1,0 +1,181 @@
+# Inventory migration guide: v1.2.0 → v2.x
+
+This document explains how to update your inventory files when migrating
+from SEAPATH ansible v1.2.0 to v2.x (main).
+
+The main breaking change is the replacement of **ceph-ansible** by an
+internal **cephadm**-based deployment, which removes the `mons`, `osds`
+and `clients` groups. The live migration user also changed.
+
+You can compare with the updated examples in this directory:
+- `seapath-cluster.yaml`
+- `seapath-standalone.yaml`
+- `seapath-vm-deployement.yaml`
+
+(`seapath-ovs.yaml` has no functional change.)
+
+---
+
+## seapath-cluster.yaml
+
+### Groups to remove
+
+Ceph is no longer deployed with ceph-ansible. Remove the following groups
+entirely:
+
+- `mons` (hosts and all its vars)
+- `osds` (hosts and all its vars)
+- `clients` (hosts and all its vars)
+
+Also remove the empty placeholder groups that existed mostly for
+ceph-ansible:
+
+```yaml
+grafana-server:
+iscsigws:
+iscsi-gws:
+mdss:
+mgrs:
+nfss:
+rbdmirrors:
+rgwloadbalancers:
+rgws:
+```
+
+### Variables to remove
+
+- `ceph_osd_disk` and `devices` (previously in the `osds` group) — replaced
+  by `ceph_osd_disks` (see below)
+- In `ceph_conf_overrides.global`:
+  - `osd_pool_default_pg_num`
+  - `osd_pool_default_pgp_num`
+
+  (PG autoscaling now handles placement group numbers)
+- Gone along with their groups: `ceph_origin`, `cluster_network`,
+  `public_network`, `monitor_address`, `configure_firewall`,
+  `ntp_service_enabled`, `dashboard_enabled`, `user_config`, `rbd`,
+  `pools`, `keys`, `ceph_keyring_permissions`
+
+### Variables to change
+
+| Variable             | Old value      | New value       |
+| -------------------- | -------------- | --------------- |
+| `livemigration_user` | `livemigration` | `libvirtadmin` |
+
+Note: to setup an observer machine, you must now remove its
+`ceph_osd_disks` variable instead of removing it from the `osds` group.
+
+### New mandatory variables
+
+Per host in `cluster_machines`, replace the old `ceph_osd_disk` of the
+`osds` group with a list of disks (one OSD per disk):
+
+```yaml
+node1:
+  # ...
+  ceph_osd_disks:
+    - "/dev/disk/by-path/pci-0000:03:00.0-scsi-0:2:1:0"
+```
+
+In `cluster_machines:vars`, add the Ceph configuration previously held by
+the `mons` group:
+
+```yaml
+  vars:
+    deploy_cephfs: false # Change to true to deploy cephfs on your cluster
+
+    # Ceph configuration
+    cephadm_network: "192.168.55.0/24" # IP range of your cluster.
+    ceph_conf_overrides:
+      global:
+        osd_pool_default_size: "{{ groups['hypervisors'] | length }}"
+        osd_pool_default_min_size: 1
+        osd_crush_chooseleaf_type: 1
+        mon_osd_min_down_reporters: 1
+      mon:
+        paxos prop interval: 100ms
+      osd:
+        osd memory target: 8076326604
+```
+
+### New optional variables
+
+- `ptp_domain_number` (in `all.vars`) — PTP domain number (0 to 255).
+  The variables `timemaster_ptp_domain_number` and
+  `ptp_status_vsock_domain_number` default to it:
+
+  ```yaml
+  ptp_domain_number: 0
+  timemaster_ptp_domain_number: "{{ ptp_domain_number }}"
+  ptp_status_vsock_domain_number: "{{ ptp_domain_number }}"
+  ```
+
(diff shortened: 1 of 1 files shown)


inventories/examples/MIGRATION-v1.2.0-to-v2.x.md
```diff
@@ -0,0 +1,181 @@
...
+    # Ceph configuration
+    cephadm_network: "192.168.55.0/24" # IP range of your cluster.
+    ceph_conf_overrides:
+      global:
+        osd_pool_default_size: "{{ groups['hypervisors'] | length }}"
+        osd_pool_default_min_size: 1
+        osd_crush_chooseleaf_type: 1
+        mon_osd_min_down_reporters: 1
+      mon:
+        paxos prop interval: 100ms
```
insatomcat, line 96: This `mon:` override does not exist anywhere in the repository, and copying this block also drops a real setting.

`git show v1.2.0:inventories/examples/seapath-cluster.yaml` and the current `inventories/examples/seapath-cluster.yaml:120` both have `auth_allow_insecure_global_id_reclaim: false` under `mon:`, and `grep -rn paxos` over the tree returns nothing.

`roles/cephadm/templates/ceph.conf.j2:13-19` renders every key of every non-global section verbatim, so a user who copies this snippet ends up with `paxos prop interval = 100ms` in the `[mon]` section of ceph.conf (not a valid Ceph option once spaces are normalised to underscores) and loses the `auth_allow_insecure_global_id_reclaim` hardening that the shipped example carries.

Suggested fix: replace the `mon:` entry with `auth_allow_insecure_global_id_reclaim: false`, matching the example inventory.

inventories/examples/MIGRATION-v1.2.0-to-v2.x.md
```diff
@@ -0,0 +1,181 @@
...
+  ```yaml
+  ptp_domain_number: 0
+  timemaster_ptp_domain_number: "{{ ptp_domain_number }}"
+  ptp_status_vsock_domain_number: "{{ ptp_domain_number }}"
+  ```
+
+- `grub_password` (in `all.vars`) — Debian hardening specific: GRUB
+  protected password. Default password is `seapath`. Generate it with
+  `grub2-mkpasswd-pbkdf2 -c 65536 / grub-mkpasswd-pbkdf2 -c 65536`.
+  If not defined, the default password is used.
```
insatomcat, line 116: "If not defined, the default password is used" is not true, and that makes `grub_password` mandatory rather than optional on Debian.

`roles/configure_hardening/defaults/main.yml` defines no `grub_password`, and `roles/configure_hardening/README.md:16` marks it `Required: Yes`. `roles/configure_hardening/tasks/main.yml:35` imports `grub.yml` unconditionally, and its first task templates `01_password.j2`, which contains a bare `{{ grub_password }}` (`roles/configure_hardening/templates/01_password.j2:4`).

A Debian user migrating from v1.2.0 (where the variable did not exist) who reads this line and skips the variable gets `seapath_setup_hardening.yaml` aborting with "'grub_password' is undefined".

The sentence about the default password refers to the hash shipped in the example being the hash of `seapath`, which is worth keeping, but the fallback claim should go and the variable should move to the mandatory section. That also applies to the standalone inventory, where line 146 currently says "New mandatory variables: None.".

inventories/examples/MIGRATION-v1.2.0-to-v2.x.md
```diff
@@ -0,0 +1,181 @@
...
+        mon_osd_min_down_reporters: 1
+      mon:
+        paxos prop interval: 100ms
+      osd:
+        osd memory target: 8076326604
+```
+
+### New optional variables
+
+- `ptp_domain_number` (in `all.vars`) — PTP domain number (0 to 255).
```
insatomcat, line 103: `timemaster_ptp_domain_number` and `ptp_status_vsock_domain_number` do not default to `ptp_domain_number`.

`roles/timemaster/defaults/main.yml:7` and `roles/ptp_status_vsock/defaults/main.yml:5` both default to the literal `0`, and nothing in either role references `ptp_domain_number`. The link exists only because the example inventory writes the two `"{{ ptp_domain_number }}"` lines by hand (`inventories/examples/seapath-cluster.yaml:41-43`).

A user who reads the prose and adds only `ptp_domain_number: 5` to `all.vars` keeps PTP domain 0 in the rendered `timemaster.conf` and in `ptpstatus.sh`, with no error to signal it.

Suggested wording: the two mapping lines are required for `ptp_domain_number` to have any effect, they are not defaults.

inventories/examples/MIGRATION-v1.2.0-to-v2.x.md
```diff
@@ -0,0 +1,181 @@
...
+
+---
+
+## seapath-vm-deployement.yaml
+
+### Groups to remove / variables to change
+
+None.
+
+### New mandatory variables
```
insatomcat, line 162: `live_migration` is not mandatory, and the instruction to add it to every VM host contradicts the updated example.

`roles/deploy_vms_cluster/tasks/main.yml:105` reads `live_migration: "{{ hostvars[item].live_migration | default(false) }}"`, so an inventory without the variable deploys normally with live migration disabled.

The updated `inventories/examples/seapath-vm-deployement.yaml` sets it only on `rtvm`, not on the second `vm` host. A user following "Per VM host, add `live_migration`" enables it for every VM, including on a standalone deployment where it does not apply.

Also, the host key used in the snippet (`seapath-vm:`) matches neither the VM inventory (`rtvm`, `vm`) nor the standalone one (`seapath-machine`).

inventories/examples/MIGRATION-v1.2.0-to-v2.x.md
```diff
@@ -0,0 +1,181 @@
...
+`osds` group with a list of disks (one OSD per disk):
+
+```yaml
+node1:
+  # ...
+  ceph_osd_disks:
+    - "/dev/disk/by-path/pci-0000:03:00.0-scsi-0:2:1:0"
+```
+
+In `cluster_machines:vars`, add the Ceph configuration previously held by
```
insatomcat, line 80: `deploy_cephfs` appears both here, inside the "New mandatory variables" block, and at line 118 as an optional variable.

The block is introduced as "the Ceph configuration previously held by the `mons` group", but `deploy_cephfs` was never in the `mons` group in v1.2.0, and `playbooks/cluster_setup_cephadm.yaml:34` guards the role with `when: deploy_cephfs is defined and deploy_cephfs | bool`, so leaving it out is fine.

Line 118 has it right. Dropping the `deploy_cephfs: false` line from this snippet would remove the contradiction and keep the block to what really moved out of `mons`.

insatomcat (commented):
Thanks for writing this up, the guide is genuinely useful and the ceph-ansible to cephadm story is accurate: the removal of the `mons`/`osds`/`clients` groups and the nine placeholder groups, the `livemigration_user` rename (commit 8c0c5cad deleted `roles/add_livemigration_user`, so a stale `livemigration` value silently breaks live migration), the `ceph_osd_disk` to `ceph_osd_disks` move with the observer note, the `osd_pool_default_pg{,p}_num` removal (commit 96371668: they make pool creation fail with `ERANGE` under the autoscaler) and the "no functional change" verdict on `seapath-ovs.yaml` all check out against `git diff v1.2.0 main -- inventories/examples/`.

I checked each claim against the roles that actually consume the variables and found five factual corrections, left inline below. Two of them (the invented `mon:` override and the `grub_password` fallback) would break or silently degrade a deployment for someone following the guide literally.

Two nits, no need to block on them:

- The document uses the em-dash character throughout (that's AI signature...). The rest of the repository documentation sticks to plain hyphens and colons.
- The two YAML snippets at lines 73-78 and 83-99 use inconsistent indentation levels (`node1:` at column 0, then `  vars:` indented by two), which makes the nesting under `cluster_machines` ambiguous for a reader copying them.

eroussy (commented):
This new inventory is not compatible with 2.0 which still needs osds groups for ceph_prepare_installation and expansion volume roles.
We should either
- write a migration guide from 2.0 to (future) 2.1
- write a migration guide from 1.2 to 2.0
Or do both, but having a 1.2 to 2.1 doesn't make sense IMO

PR #1012, opened 2026-08-29
prepare: check jmespath dependency

joyshmitz:
The documented non-cqfd setup lists `netaddr` and `six`, but several roles use `community.general.json_query`, which requires the `jmespath` Python module in the controller's environment.

With the ansible repository at v2.0.1, ansible-core 2.16, and a clean Python environment containing the documented dependencies but not jmespath, `seapath_setup_main.yaml` failed at `roles/cephadm/tasks/main.yml` after Ceph bootstrap:

```
You need to install "jmespath" prior to running json_query filter
```

The cqfd image already installs jmespath (`.cqfd/docker/Dockerfile`), and the virtual-cluster CI workflow pip-installs it before running `prepare.sh` — only the native (non-cqfd) path is missing it.

This PR:
- adds a jmespath check to `prepare.sh`, following the existing netaddr check style, so the failure happens early with an actionable message;
- mentions the prerequisite in `README.adoc`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -9,7 +9,7 @@
   cqfd init           # build the dev container
   cqfd -b prepare     # install galaxy deps, submodules, patches, plugins
   ```
-- **Without cqfd**: install `ansible-core~=2.16.0`, `netaddr`, `six`, `rsync`, then run `./prepare.sh`.
+- **Without cqfd**: install `ansible-core~=2.16.0`, `netaddr`, `six`, `jmespath`, `rsync`, then run `./prepare.sh`.
 
 ## Lint / Format
 
--- a/README.adoc
+++ b/README.adoc
@@ -122,7 +122,7 @@ https://docs.ansible.com/ansible/latest/installation_guide/intro_installation.ht
 *Warning:* Currently only the Ansible version 2.16 is supported. Other versions
 will not work.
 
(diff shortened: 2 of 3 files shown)


insatomcat:
The check itself looks good: it mirrors the existing `netaddr` block, runs before `ansible-galaxy install`, and the requirement is real (`community.general.json_query` is used in `roles/cephadm/tasks/main.yml`, lines 223 and 307). The cqfd path stays fine since `.cqfd/docker/Dockerfile` already pip-installs `jmespath` into `/env`, which is first in `PATH`.

One thing to amend: `AGENTS.md` line 12 duplicates the non-cqfd dependency list ("install `ansible-core~=2.16.0`, `netaddr`, `six`, `rsync`, then run `./prepare.sh`") and was not updated along with `README.adoc`. Following it now gives a clean environment that fails on the new check. Could you add `jmespath` there too, so both documents agree?

Optional, while you are in that line of `README.adoc`: it reads "Also you must also install", which duplicates "also". The PR rewrites that exact line, so it is a free fix.

joyshmitz:
Thanks for the review. Addressed in bb63d0e: `AGENTS.md` now lists `jmespath` in the non-cqfd dependency line, and the README sentence reads "You must also install …" (duplicated "also" removed). The `prepare.sh` check itself is unchanged.

insatomcat (approved):

PR #1014, opened 2026-09-02
cephadm: apply ceph_conf_overrides to a running cluster

insatomcat:
`ceph_conf_overrides` only reaches the cluster through the `ceph.conf` handed to `cephadm bootstrap --config`, and that template is uploaded only when `cephadm_do_bootstrap` is true. On a cluster that is already bootstrapped, editing `ceph_conf_overrides` in the inventory and re-running the playbook changes nothing and reports nothing.

The role now writes the options to the cluster configuration store as well. It reads the running value of each option with `ceph config get` and calls `ceph config set` only on those that differ, so a converged cluster reports no change. Option names are normalised to underscores, the way `ceph config` spells them.

Two effects worth knowing: an option Ceph does not know about now fails the run, where the `ceph.conf` parser dropped it silently; and removing an option from `ceph_conf_overrides` leaves its value in the cluster, as the role never unsets anything (`ceph config rm` goes back to the default).

Picks up one of the valid points of #867, which will be closed with a reference to this PR.


insatomcat:
The `ci-virtual-cluster` job fails on `ceph config set global pg_autoscale_mode on`, with Ceph rejecting the option name.

`pg_autoscale_mode` is a pool property. The matching cluster-wide configuration option is `osd_pool_default_pg_autoscale_mode`. It comes from the CI inventory in `seapath/virtual-cluster` (`inventory/seapath-sandbox.yaml`), where the `ceph.conf` parser has been dropping it in silence, which is exactly what this PR stops doing. seapath/virtual-cluster#11 fixes the option name; this PR needs it merged to go green.

The other six options of that inventory were checked against the upstream Ceph option list and are all valid.

eroussy (dismissed):

dupremathieu (approved):

PR #1015, opened 2026-09-02
cluster_remove_machine: refuse to drop a machine the cluster has not lost

insatomcat:
`cluster_remove_machine.yaml` removes the Ceph host with `--force --offline`, so the host is dropped without being drained and its OSDs are abandoned. That is the right move for a machine the cluster has already lost, which is what this playbook is for, and nothing checks that this is the situation. Aimed at a healthy machine, or at the wrong one after a typo in `machine_to_remove`, it leaves the cluster degraded without a word.

The playbook now reads `ceph osd tree` first and stops if the machine still has OSDs in the `up` state, pointing at `ceph orch host drain` as the way to retire a machine that is alive. A machine whose OSDs are all down, or that carries no OSD, is removed as before.

Asking Ceph rather than probing the machine over SSH keeps the answer meaningful when the control node reaches the cluster through a bastion, and it asks the question that matters: would this removal abandon data that is still being served.

It also refuses a `machine_to_remove` absent from `cluster_machines`, which used to fail later on an undefined hostvars lookup.

Picks up one of the valid points of #867, which will be closed with a reference to this PR.

---

A second commit applies the same reasoning to Pacemaker, which the playbook removes the machine from without ever asking it anything.

`crm_node -R` returns a success against a machine that is still an active Corosync member, and the removal is then silently lost. Measured on a three node bench, removing a live node: the command reported `changed`, the node stayed `member` in `crm_node -l`, Corosync and Pacemaker were still running on it, and its `ceph-mgr` was still the active manager of the cluster it had just been removed from.

The playbook now reads `crm_node -l` before acting and stops if the machine is still an active member, symmetrically to the OSD guard. A machine that is powered off shows up as `lost` and is removed as before. It reads the node list again at the end and fails if the machine is still listed, since this is the command that lies.

The two guard blocks were exercised on a three node bench: refusal against a live member, and a clean run against a powered off one.

### DIFF
--- a/playbooks/cluster_remove_machine.yaml
+++ b/playbooks/cluster_remove_machine.yaml
@@ -61,6 +61,30 @@
           playbook again.
       run_once: true
 
+    # crm_node -R returns a success against a machine that is still an active
+    # Corosync member and the removal is then silently lost: the node stays in
+    # the member list and its ceph-mgr keeps running on a cluster it no longer
+    # belongs to. Ask Pacemaker the same way we ask Ceph.
+    - name: Get the Pacemaker node list # noqa: run-once[task]
+      ansible.builtin.command: crm_node -l
+      register: pacemaker_node_list
+      changed_when: false
+      delegate_to: "{{ first_node }}"
+      run_once: true
+
+    - name: Refuse to remove a machine still active in the Corosync ring # noqa: run-once[task]
+      ansible.builtin.assert:
+        that: >-
+          pacemaker_node_list.stdout_lines
+          | select('search', '\s' ~ machine_to_remove ~ '\s+member$')
+          | list | length == 0
+        fail_msg: >-
+          {{ machine_to_remove }} is still an active Corosync member. Removing
+          it now would be reported as a success and silently lost, and its
+          ceph-mgr would keep running on a cluster it no longer belongs to.
+          Power {{ machine_to_remove }} off, then run this playbook again.
+      run_once: true
+
     - name: Remove machine from pacemaker-corosync cluster # noqa: run-once[task]
       ansible.builtin.command: "crm_node -R {{ machine_to_remove }} --force"
       delegate_to: "{{ first_node }}"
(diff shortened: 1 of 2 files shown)


playbooks/cluster_remove_machine.yaml
```diff
@@ -20,6 +25,42 @@
...
+      run_once: true
+
+    - name: Refuse to remove a machine whose OSDs are still up # noqa: run-once[task]
+      ansible.builtin.assert:
+        that: machine_to_remove_up_osds | length == 0
+        fail_msg: >-
+          {{ machine_to_remove }} still has OSDs up in the cluster
+          ({{ machine_to_remove_up_osds | join(', ') }}). Removing it now would
+          abandon them and leave the cluster degraded. Drain the machine first
+          with "ceph orch host drain {{ machine_to_remove_hostname }}", wait for
```
eroussy, line 59: Shouldn't the playbook handle the case where we want to drain the machine ?
It seems we have two scenarios here:
- one where the machine is already powered off and we simply remove it from ceph known osds 
- one where the machine is still part of the cluster and we want to remove it "correctly" with all the proper ceph commands

Should we make two different cases for this playbook then ?

eroussy (changes requested):

insatomcat:
Yes, these are two different operations, and this PR deliberately keeps cluster_remove_machine.yaml to the first one (a machine the cluster has lost). The guard is there so that the second case is refused instead of silently degrading the cluster.
This playbook was never about the second case, so I would rather not add a drain mode here, because the two paths share almost nothing:
- Ceph: with the default inventory (osd_pool_default_size = number of hypervisors, host failure domain), ceph orch host drain can never complete, since PGs have no other host to move to. A clean removal first needs the pools' size lowered, which is a replication policy decision on its own.
- Pacemaker: crm_node -R against a live member is silently lost (measured on a bench). A clean removal needs standby, resource migration, then stopping corosync/pacemaker on the machine itself before removing it.

if decommissioning a live machine is a need, it should have its own playbook and PR.

PR #1016, opened 2026-09-05
cyclictest: add a standalone playbook and check for the binary

insatomcat:
The `cyclictest` role is only reachable through `ci_all_machines_tests.yaml`, which runs it after the Yocto functional tests. Measuring latency on a Debian machine, or on a deployment that is already running, means editing that playbook or calling the role by hand.

The first commit adds `playbooks/test_run_cyclictest.yaml`: the role on its own, against `cluster_machines` and `standalone_machine`, the groups the other `seapath_setup` playbooks play. Nothing on the targets changes. The role copies a script to a temporary directory, runs cyclictest, fetches the histogram to the control machine and leaves.

The second commit makes a missing binary readable. cyclictest comes from the `rt-tests` package and nothing in this collection installs it, so on a machine without it the role reached the script, which does not catch the missing binary, and the operator got a Python traceback ending in `FileNotFoundError`. The role now looks for the binary first and fails naming the package. It still installs nothing: a latency measurement must change nothing on the machines it measures. The README claimed "no requirement", which was wrong.

### DIFF
--- a/playbooks/test_run_cyclictest.yaml
+++ b/playbooks/test_run_cyclictest.yaml
@@ -1,0 +1,24 @@
+# Copyright (C) 2026, RTE (http://www.rte-france.com)
+# SPDX-License-Identifier: Apache-2.0
+
+# Ansible playbook that measures the latency of the hypervisors with
+# cyclictest and fetches the histogram to the control machine.
+#
+# The `cyclictest` role has only ever been reachable through
+# `ci_all_machines_tests.yaml`, which runs it after the Yocto functional tests
+# and therefore cannot be used on a Debian machine or on a running deployment.
+# This playbook is that role on its own, so a latency measurement is something
+# an operator can ask for on any SEAPATH machine.
+#
+# Nothing on the machines is changed: the role copies a script to a temporary
+# directory, runs cyclictest, fetches the result and leaves.
+
+---
+- name: Measure the latency with cyclictest
+  hosts:
+    - cluster_machines
+    - standalone_machine
+  become: true
+  gather_facts: false
+  roles:
+    - cyclictest
--- a/roles/cyclictest/README.md
+++ b/roles/cyclictest/README.md
@@ -4,7 +4,11 @@ This role runs cyclictest and fetch the resulting histogram.
 
 ## Requirements
 
-no requirement.
+`cyclictest`, from the `rt-tests` package. Nothing in this collection installs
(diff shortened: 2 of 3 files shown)


eroussy (changes requested):
That's great. I would go further and use that playbook during the Yocto and SLES CI too.
Two options:
- Replace cyclictest role call in `ci_all_machine_tests` by this new playbook
- Completely remove cyclictest call in `ci_all_machine_tests` and add this new playbook in the `hyp-test-playbooks` section of both yocto and sles action files. Then rename `ci_all_machine_tests` into `ci_launch_cukinia_test` (We should also harmonize that later though)

I would go for the second option.

insatomcat:
Thanks, glad it is useful. I would rather keep that out of this PR though, if you don't mind: I am not the one running the Yocto or SLES benches, so the call on which option to take is yours, and the current PR is deliberately limited to making the role usable on its own and readable when the binary is missing, with no change to what the CI does today.

Switching the CI over is also a bit more involved than moving a playbook name around. Three things I ran into while looking at it, so you have them when you decide:

1. `skip_cyclictest`: The `when: skip_cyclictest is not defined or not skip_cyclictest` in `ci_all_machines_tests.yaml` is evaluated per host, and the private inventories set the variable per machine (that is what the comment in `ci-debian.yml` refers to). `test_run_cyclictest.yaml` has no such guard, so dropped into `hyp-test-playbooks` as it stands it would measure the hosts that the inventory marks as skipped, and `.github/test-report/cyclictest.adoc.j2` then filters those results out of the report. The condition needs to move into the new playbook.

2. The VMs. `ci_all_machines_tests.yaml` is used both as `hyp-test-playbooks` and as `vms-test-playbooks`, and its `hosts` include `VMs`, so cyclictest currently also runs on the RT VMs and the report has a "Virtual Machines" section that depends on it. Adding the new playbook only on the hypervisor side would drop that measurement. Either the playbook targets `VMs` as well, or it is listed in both inputs.

3. The rename:`ci_all_machines_tests` appears four times in `ci-yocto.yml` and `ci-sles.yml`, and once in `README-ansible-galaxy.md`, where the playbook is part of the published collection. Renaming a public playbook feels like it deserves its own commit or its own PR, along the harmonisation you mention.

So my suggestion would be a follow-up PR from you once you have settled on option 1 or 2.

PR #1017, opened 2026-09-05
feat: add a hwlatdetect role and its playbook

insatomcat:
`hwlatdetect` reports the interruptions the kernel never sees. An SMI takes the CPU into firmware without telling the operating system, so the time is missing from the kernel's own accounting and from cyclictest's view of it. A machine that is correctly isolated and still misses its deadline is either a firmware problem or a configuration one, and this is the measurement that tells the two apart.

The role runs the detector and fetches the result, plus `playbooks/test_run_hwlatdetect.yaml` to play it against `cluster_machines` and `standalone_machine`. Nothing here installs `rt-tests`, and nothing on the targets changes.

Both missing prerequisites fail the play before any machine starts measuring: a missing `hwlatdetect` binary names the `rt-tests` package, a kernel without the hwlat tracer names `CONFIG_HWLAT_TRACER`. A run that does not complete fails the play too, but only after the results are fetched, since the machines measure in parallel and `any_errors_fatal` would otherwise drop the results of the machines that did measure. The exit status alone cannot tell a failure: hwlatdetect exits 1 when the worst gap exceeds the threshold, which is a finding, and a crash exits 1 as well. The role fails when the exit status is above 1 or when the report line is missing from stdout. The fetched file holds the exit status, stdout and stderr.

Run it on a machine carrying no live traffic. The hwlat thread moves to another CPU at the start of every window, round robin over `tracing_cpumask`, so the isolated cores are sampled like the rest. That is the point, the SMI worth finding is the one that hits an isolated core, and it is what the measurement costs: the core the thread lands on runs `width` microseconds with interrupts disabled, 0.5s at the defaults. No guest processing Sampled Values survives that, which is why the defaults are documented as a bench setting and why the role stays out of `seapath_setup_main.yaml`.

### DIFF
--- a/playbooks/test_run_hwlatdetect.yaml
+++ b/playbooks/test_run_hwlatdetect.yaml
@@ -1,0 +1,29 @@
+# Copyright (C) 2026, RTE (http://www.rte-france.com)
+# SPDX-License-Identifier: Apache-2.0
+
+# Ansible playbook that measures the interruptions the kernel cannot see, with
+# hwlatdetect, and fetches the result to the control machine.
+#
+# This is the measurement no other check on a SEAPATH machine can make. A
+# System Management Interrupt takes the CPU into firmware without telling the
+# operating system, so the time is missing from the kernel's accounting and
+# from cyclictest's view of it. A correctly isolated machine that still misses
+# its deadline is either a firmware problem or a configuration one, and this is
+# what tells the two apart.
+#
+# Nothing on the machines is changed. Run it where no live traffic is carried,
+# on a bench, on a machine not yet in service, or on one taken out of the
+# cluster: the tracer rotates over every CPU, isolated cores included, and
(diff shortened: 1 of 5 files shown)


eroussy (changes requested):
Seems weird to me to write the failing reason of hwlatdetect in its result files. I would go into letting the ansible task fail if hwlatdetect fails.
Rest of the code looks fine to me

insatomcat:
yes you're right, the thing is hwlatdetect exits 1 whenever it finds a latency above the threshold, the same code as a crash, so the exit code alone cannot tell a failure from a finding...
So I changed to: 
- a missing tracer fails before any machine starts measuring
- a run fails the play when the exit code is above 1 or when the report line is missing from the output
- this check comes after the fetch so the other machines keep their results

